PR #7, opened 2021-06-23
App crashes when attempting to access table of contents

mauricecarrier7:
**What's this do?**
Resolves crash caused by attempting to access table of contents

Updates the TPPReaderTOC.storyboard viewController identifier to TPPReaderTOC
Updates the call sites to instantiate the storyboard and view controller to use the correct identifiers.

Attempt to access table of contents from the application, app should not crash

No

Yes, @mauricecarrier7


vladimirfedorov:
It still crashes if the app uses R2 reader, R2 uses another view controller, `TPPReaderPositionsVC`, to show TOC.

mauricecarrier7:
> It still crashes if the app uses R2 reader, R2 uses another view controller, `TPPReaderPositionsVC`, to show TOC.

Ah, missed that path, updated.

vladimirfedorov (approved):

PR #9, opened 2021-06-25
Add Search to account selection screen

mauricecarrier7:
**What's this do?**
- Implements search functionality for account selection
- refactors account selection screen
- displays logos on account selection screen

- improve user experience

- access account selection screen
- attempt to use search functionality

N/A

No

N/A

@mauricecarrier7 

https://user-images.githubusercontent.com/6921353/123489004-17e15a00-d5df-11eb-85a8-0e7732c6dedd.mov


Palace/Settings/AccountList/TPPAccountListDataSource.swift
```diff
@@ -0,0 +1,60 @@
...
+  override init() {
+    super.init()
+    loadData()
+  }
+  
+  private func loadData(_ filterString: String? = nil) {
+    accounts = AccountsManager.shared.accounts()
+    accounts.sort { $0.name < $1.name }
+    
+    if let filter = filterString, !filter.isEmpty {
```
vladimirfedorov, line 24: Do you think we can use case-insensitive search here? Would be easier to find a library when the user doesn't know its exact spelling.

mauricecarrier7: Nice catch, I missed that due to the search bar capitalizing the first string by default. Updated

vladimirfedorov (approved):
Looks good to me

PR #19, opened 2021-07-09
Implement dark mode support

mauricecarrier7:
**What's this do?**
Updates UI elements to support dark mode

https://www.notion.so/lyrasis/Highlighted-Icons-on-iOS-too-dark-923f6d227bf644ab9a4cdcee30e00c9f

Place device in dark mode and use the application

N/A

N/A

N/A

@mauricecarrier7 

https://user-images.githubusercontent.com/6921353/125130255-afaa7200-e0ce-11eb-8cde-c8e6cbf8e51f.mov


Palace/AppInfrastructure/TPPConfiguration+SE.swift
```diff
@@ -44,10 +44,14 @@ extension TPPConfiguration {
...
-    return UIColor(red: 141.0/255.0, green: 199.0/255.0, blue: 64.0/255.0, alpha: 1.0)
+    UIColor(red: 141.0/255.0, green: 199.0/255.0, blue: 64.0/255.0, alpha: 1.0)
   }
 
   static func cardCreationEnabled() -> Bool {
     return true
   }
+  
+  @objc static func iconColor() -> UIColor {
+    UIApplication.darkModeEnabled ? .white : .black
```
vladimirfedorov, line 55: I think we should use either system colors for iOS 13+ (e.g., UIColor.label) or create an asset color: right now the app doesn't change button colors when iOS changes dark mode value, rendering buttons invisible:
Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-07-13 at 00 10 15

Palace/Utilities/UI/UIApplication+Extensions.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+@objc extension UIApplication {
+  static var darkModeEnabled: Bool {
+    if #available(iOS 13, *) {
+      return UITraitCollection.current.userInterfaceStyle == .dark
```
vladimirfedorov, line 6: Probably we can avoid this global variable by adding a color, compatible with iOS versions prior to 13, to UIColor, something like `compatibleTextColor` (just an example) that returns .label/asset color for iOS 13+ and black for other versions, as we are not really interested in dark mode value itself.

Palace/AppInfrastructure/TPPAppTheme.swift
```diff
@@ -25,7 +25,7 @@
       return colorFromHex(hex(theme))
```
vladimirfedorov, line 25: Do you think we can use available system colors for iOS 13+ here for matching values? they are already optimised for dark mode, and change automatically when iOS changes dark mode value.

Palace/AppInfrastructure/TPPAppTheme.swift
```diff
@@ -25,7 +25,7 @@
...
     } else {
       Log.error(#file, "Given theme color is not supported: \(name)")
-      return .black
+      return UIApplication.darkModeEnabled ? .white : .black
```
vladimirfedorov, line 28: (added description for iconColor below) when iOS changes dark mode value, text remains the same color it was set when the app started.

Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-07-13 at 00 54 30

vladimirfedorov (changes requested):

Palace/AppInfrastructure/TPPConfiguration+SE.swift
mauricecarrier7: Good call, updated.

mauricecarrier7: Ah I see now, this is going to be a bit tricker since the reader can be in a different mode than the app itself.

vladimirfedorov (approved):

PR #20, opened 2021-07-12
Disable copy and paste in app

mauricecarrier7:
**What's this do?**
Disables copy and paste for iOS 11 and above

https://www.notion.so/lyrasis/Disable-Copy-Paste-in-iOS-App-33f4256072f742e59221d8f0591647ad

Attempt to copy and paste text from an e-book

N/A

N/A

N/A 

@mauricecarrier7


vladimirfedorov (changes requested):
I don't think it works as described in the ticket. This option works in R1 only, and it removes "Copy" button in the strip above the selected text. Still, the text can be selected and copied via "Share" sheet, there's another "Copy" action in that sheet. Also, this doesn't affect R2, it shows all available actions for selected text.

Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-07-13 at 18 33 43

vladimirfedorov (approved):

PR #28, opened 2021-07-19
Default to Readium 2

mauricecarrier7:
**What's this do?**
Defaults application to use Readium 2 and allows R1 to be enabled from the developer settings menu

Readium 1 is being deprecated

View an ebook, observe default reader is R2. Go to settings and toggle on R1, view an ebook, observe reader is R1.

N/A

N/A


Palace.xcodeproj/project.pbxproj
```diff
@@ -2491,7 +2480,7 @@
 		A823D804192BABA400B55DE2 = {
 			isa = PBXGroup;
 			children = (
-				E585662026979AC100A5FBD5 /* NYPLAEToolkit.xcodeproj */,
+				E585662026979AC100A5FBD5 /* ios-drm-audioengine */,
```
vladimirfedorov, line 2483: What is the benefit of removing NYPLAEToolkit project and replacing it with the folder? From what I see, Xcode won't build it and any change in the toolkit won't take effect until Carthage rebuilds it.

vladimirfedorov (changes requested):

PR #31, opened 2021-07-22
Default to Readium 2

mauricecarrier7:
What's this do?
Defaults application to use Readium 2 and allows R1 to be enabled from the developer settings menu

Why are we doing this? (w/ JIRA link if applicable)
Readium 1 is being deprecated

How should this be tested? / Do these changes have associated tests?
View an ebook, observe default reader is R2. Go to settings and toggle on R1, view an ebook, observe reader is R1.

Dependencies for merging? Releasing to production?
N/A

Has the application documentation been updated for these changes?
N/A

Did someone actually run this code to verify it works?
@mauricecarrier7

### DIFF
--- a/Palace/Book/UI/TPPBookCellDelegate.m
+++ b/Palace/Book/UI/TPPBookCellDelegate.m
@@ -130,20 +130,20 @@
 
 - (void)openEPUB:(TPPBook *)book
 {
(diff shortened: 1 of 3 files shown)


vladimirfedorov (approved):
Looks good to me

PR #32, opened 2021-07-22
Remove Readium 1

mauricecarrier7:
What's this do?
Removes source strictly related to the R1 reader

Why are we doing this? (w/ JIRA link if applicable)
Remove deprecated code from the code base

How should this be tested? / Do these changes have associated tests?
Attempt to view/read an ebook, Readium 2 should be used by default. The reader version toggle option should not be present in the developers menu.

Dependencies for merging? Releasing to production?
N/A

Has the application documentation been updated for these changes?
N/A

Did someone actually run this code to verify it works?
@mauricecarrier7


vladimirfedorov (changes requested):
That's a good start, we should be able to remove code in `Reader` folder (it was for R1) and `Adobe Content Filter` as well (also needed for R1 only).

vladimirfedorov (approved):

PR #37, opened 2021-07-30
Replace NYPL submodules with ThePalaceProject ones

vladimirfedorov:
**What's this do?**
This PR removes NYPL references from Palace code:
- replaces NYPL submodules with ThePalaceProject submodules;
- removes SimplyE keychain migration;
- removes OpenEbooks references in the code.

https://www.notion.so/lyrasis/Remove-NYPL-References-8d73b02de51e41dc8b448e13160b70cc

As this PR updates git submodules, it is better to do clean setup to avoid any interference with old submodule folders. The project shouldn't use NYPL repository and shouldn't have "org.nypl" references in code, e.g., in keychain names.

No

No

Yes (README references in bookmark tests)

@vladimirfedorov


mauricecarrier7 (approved):
I was able to successfully build and run this update. Looks good to me.

PR #39, opened 2021-08-05
Custom library registry fails to load

mauricecarrier7:
**What's this do?**
Ensures that catalog is loaded when a custom registry is added. Enable the ability to add multiple custom libraries

https://www.notion.so/lyrasis/Unable-to-configure-alternate-library-registry-in-iOS-app-8873852fdc824e148ef3c2c2bbfbdd1d

Go to developer settings, add custom registry attempt to add library from registry

N/A

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good

PR #40, opened 2021-08-05
Prevent infinite loop

mauricecarrier7:
**What's this do?**
Modifies custom registry reset functionality to prevent attempt to reset account after cache is cleared. Adds restrictions to registry input field.

The logic path that initally supported custom registry reset caused an infinite loop when calling `updateAccountSetFromNotification`
 
Run the app attached to the debugger and see if extraneous network calls are made.

N/A

N/A

N/A

@mauricecarrier7

### DIFF
--- a/Palace/Accounts/Library/AccountsManager.swift
+++ b/Palace/Accounts/Library/AccountsManager.swift
@@ -304,9 +304,7 @@ let currentAccountIdentifierKey  = "TPPCurrentAccountIdentifier"
 
     if self.accounts().isEmpty || TPPConfiguration.customUrlHash() != nil {
       loadCatalogs(completion: completion)
-    } else {
-      completion?(true) //Called when resetting a custom registry
-    }
+    } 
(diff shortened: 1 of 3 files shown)


Palace/Accounts/Library/AccountsManager.swift
```diff
@@ -304,9 +304,7 @@ let currentAccountIdentifierKey  = "TPPCurrentAccountIdentifier"
 
     if self.accounts().isEmpty || TPPConfiguration.customUrlHash() != nil {
       loadCatalogs(completion: completion)
-    } else {
-      completion?(true) //Called when resetting a custom registry
```
mauricecarrier7, line 308: This was being called when setting updating accounts, even for normal accounts, and was triggering an infinite loop. No longer need this line because we won't attempt to reset the account after clearing a custom registry.

vladimirfedorov (approved):

PR #45, opened 2021-08-12
Update color of audiobook icon

mauricecarrier7:
**What's this do?**
Sets audiobook icon to white for both dark and light mode

 https://www.notion.so/lyrasis/Update-iOS-UI-and-implement-new-Material-Icons-in-iOS-to-synch-up-with-Android-UI-8583437e2df24b9aa05fb4fcd34f0484

View an audiobook title

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
It looks fine, but could you change `feature/palace-init` to `develop`? We shouldn't merge anything to `palace-init` anymore, I kept it in case you had some work in progress on it but we should really delete it now.

mauricecarrier7:
> It looks fine, but could you change `feature/palace-init` to `develop`? We shouldn't merge anything to `palace-init` anymore, I kept it in case you had some work in progress on it but we should really delete it now.

The previous changes for the UI story also went into palace/init, so I will be merging those into develop with this merge as well.

PR #51, opened 2021-08-17
Reenable TPPButtonsView on state change and update the buttons state on return to detail view

mauricecarrier7:
**What's this do?**
- Reenables detail view buttons when returning to the detail view
- Updates font on empty screen

https://www.notion.so/lyrasis/Read-and-Delete-buttons-are-not-clickable-after-going-back-to-the-previous-screen-0e7b511f19cc4e2280553cb62e31f15d

Select read or listen from the detail book view, navigate back to detail view, attempt to select read or listen again.

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me

PR #54, opened 2021-08-20
Don't add libraries that already are on the list

vladimirfedorov:
**What's this do?**
- Checks if selected library already is on the list of libraries

It is possible to add libraries that are already on the list ([notion](https://www.notion.so/lyrasis/There-is-a-possibility-to-add-multiple-count-of-DPLA-libraries-5fdb618d71694403989850dec03750d7))

Please follow "Steps to reproduce" in the ticket.

NA

**Does this include changes that require a new palace build for QA?**
Yes

Yes

@vladimirfedorov

### DIFF
--- a/Palace/Settings/TPPSettingsAccountsList.swift
+++ b/Palace/Settings/TPPSettingsAccountsList.swift
@@ -176,7 +176,6 @@
     }
     updateSettingsAccountList()
     // Return from search screen to the list of libraries
-    updateNavBar()
     tableView.reloadData()
     navigationController?.popViewController(animated: false)
     // Switch to the selected library
--- a/Palace/Settings/TPPSettingsAccountsList.swift
+++ b/Palace/Settings/TPPSettingsAccountsList.swift
@@ -176,7 +176,6 @@
     }
     updateSettingsAccountList()
     // Return from search screen to the list of libraries
-    tableView.reloadData()
     navigationController?.popViewController(animated: false)
     // Switch to the selected library
     AccountsManager.shared.currentAccount = account
--- a/Palace/Settings/TPPSettingsAccountsList.swift
+++ b/Palace/Settings/TPPSettingsAccountsList.swift
@@ -171,7 +171,9 @@
   }
   
   private func updateList(withAccount uuid: String) {
-    userAddedSecondaryAccounts.append(uuid)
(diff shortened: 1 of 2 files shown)


Palace/Settings/TPPSettingsAccountsList.swift
```diff
@@ -171,8 +171,15 @@
   }
   
   private func updateList(withAccount account: Account) {
-    userAddedSecondaryAccounts.append(account.uuid)
+    if !userAddedSecondaryAccounts.contains(account.uuid) {
+      userAddedSecondaryAccounts.append(account.uuid)
+    }
     updateSettingsAccountList()
+    // Return from search screen to the list of libraries
+    updateNavBar()
```
mauricecarrier7, line 179: Shouldn't need this anymore since we are jumping directly to the catalog after adding a new account
```suggestion
```

Palace/Settings/TPPSettingsAccountsList.swift
```diff
@@ -171,8 +171,15 @@
...
+    tableView.reloadData()
```
mauricecarrier7, line 180: same as above
```suggestion
```

Palace/Settings/TPPSettingsAccountsList.swift
```diff
@@ -171,8 +171,15 @@
...
+    navigationController?.popViewController(animated: false)
```
mauricecarrier7, line 181: this is now covered in line 187 below
```suggestion
```

mauricecarrier7 (changes requested):
Looks good, but my recent PR removes the need for a couple of lines.

Palace/Settings/TPPSettingsAccountsList.swift
vladimirfedorov: it is not, "Settings" screen remains on "Pick Your Library" screen:

https://user-images.githubusercontent.com/941531/130286360-e7823bcd-fcdd-49ce-bd8e-6b710e53d9cf.mov

mauricecarrier7: Ah, forgot to check iPad.

mauricecarrier7 (approved):

PR #71, opened 2021-10-04
Disable analytics on non production builds

mauricecarrier7:
**What's this do?**
Disables Analytics for non-production builds

https://www.notion.so/lyrasis/Automated-tests-are-causing-usage-statistics-in-firebase-to-be-incorrect-iOS-1762b8f79d12492289b326d20cb943f7

Use TestFlight, Adhoc or Debug build and check if interactions are reported to Crashlytics

N/A

Yes

N/A

@mauricecarrier7  verified that the fix works locally, will need to be uploaded to BrowserStack using the current process to verify that workflow as well.


vladimirfedorov (approved):
Looks good to me

PR #81, opened 2021-11-02
Update TPPBookButtonsView.m

mauricecarrier7:
**What's this do?**
Enables users to return un-downloaded books

https://www.notion.so/lyrasis/iOS-Add-ability-to-return-a-book-before-downloading-d2c8d96da11449bea0accfb1aeabb3e5

Checkout a book on a device,  log into the same account on another device and attempt to return the title without downloading it.

N/A

yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me

PR #97, opened 2021-12-01
Fix/drop pre iOS13 support

mauricecarrier7:
**What's this do?**
Bumps deployment target to iOS 13

This change will support the modernization efforts of the application. Also relates to resolving https://www.notion.so/lyrasis/Search-bar-is-hidden-when-adding-a-new-library-c1c9b7bb995748e8bf0d59b858b09546

Regular app usage will verify this change. App will not be available on devices running iOS 12 and lower

N/A

yes

N/A

@mauricecarrier7 

Merge after: https://github.com/ThePalaceProject/ios-core/pull/95


vladimirfedorov (changes requested):

vladimirfedorov (approved):
Looks good to me

PR #109, opened 2021-12-14
Remove delay in reporting of last read position

mauricecarrier7:
**What's this do?**
Reduces the delay on the throttle mechanism used to restrict how often reading position is reported to the server.

[[Quick blurb about why the code is needed and Notion link goes here / Do these changes meet the business requirements of the story?]
](https://www.notion.so/lyrasis/Palace-does-not-remember-place-in-book-or-audiobook-between-devices-on-iOS-4e408de1eb1e49d49da3853ca957e4f4)

Enable sync on two devices, download and read a book. Open the book on another device, book should sync to the last page previously read on the other device.

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me

PR #110, opened 2021-12-15
Support Octet Stream content type

mauricecarrier7:
**What's this do?**
Extends support for octet stream content type, the type returned for the title "The Mirror of Kong Ho".

https://www.notion.so/lyrasis/App-crashes-after-getting-the-book-iPhone8-742898f53e2547efa4c6f5d43296b816

Navigate through catalog to a book, Palace Bookshelf -> Fiction -> Historical Fiction -> The Mirror of Kong Ho -> Get
- app should not crash

N/A

Yes

N/A

@mauricecarrier7


Palace/Book/Models/TPPBookContentType.m
```diff
@@ -5,7 +5,7 @@ TPPBookContentType TPPBookContentTypeFromMIMEType(NSString *const string)
 {
   if ([[TPPOPDSAcquisitionPath audiobookTypes] containsObject:string]) {
     return TPPBookContentTypeAudiobook;
-  } else if ([string isEqualToString:ContentTypeEpubZip]) {
+  } else if ([string isEqualToString:ContentTypeEpubZip] || [string isEqualToString:ContentTypeOctetStream]) {
```
mauricecarrier7, line 8: @vladimirfedorov, This fix effectively resolves the issue, however, I am still unsure why we are only seeing certain titles in this content mode and only when accessing them by navigating through the catalog. Any ideas?

mauricecarrier7: This issue actually is not related to the method of accessing a title. I tracked the issue to the title "The Mirror of Kong Ho" which is actually triggering the crash (TPPBookButtonsView.updateButtons` line 213) due to it's content-mode being the unsupported type `octet-stream`.

PR #133, opened 2022-03-03
Modifications to support cross platform sync

mauricecarrier7:
**What's this do?**
Minor modifications to support annotation syncing with Android

https://www.notion.so/lyrasis/Bookmarks-and-reading-position-are-not-synced-between-iOS-and-Android-devices-07748eff6f6b46cbb04d1837a925329d

Set a bookmark on an android/iOS device and validate position on the opposite device.

N/A

yes

N/A

@mauricecarrier7


Palace/Reader2/Bookmarks/TPPBookmarkSpec.swift
```diff
@@ -171,7 +185,11 @@ struct TPPBookmarkSpec {
        bookID: String,
        selectorValue: String) {
     self.id = Id(value: id)
-    self.body = Body(time: time.rfc3339String(), device: device)
+    
+    let dict = try? JSONSerialization.jsonObject(with: selectorValue.data(using: .utf8)!, options: []) as? [String: Any]
+    let title = (dict!!["title"] as? String) ?? ""
```
vladimirfedorov, line 190: Maybe we can avoid double force-unwrapping here? Don't know if `selectorValue` can be an empty string `""`, but the app will crash in this case.

vladimirfedorov (changes requested):

Palace/Reader2/Bookmarks/TPPBookmarkSpec.swift
mauricecarrier7: Nice catch, I meant to go back and fix this, it was for debugging purposes only. Will update

PR #150, opened 2022-04-19
Fix PDF hide/show navigation

vladimirfedorov:
**What's this do?**
This makes navigation bars hide/show behaviour similar between iOS 15 and earlier versions.

iOS 15 recognizes taps on PDFView itself, while iOS versions before 15 don't register these taps. This change also removes the additional tap gesture recognizer added to the PDFView parent view to hide/show navigation bars ([Ticket](https://www.notion.so/lyrasis/Can-t-activate-navigation-bar-for-PDFs-on-iOS-13-and-14-3cb5c44bca3a4d24b7c351fe34037275)).

In "Palace Bookshelf" open "Active Calculus Multivariable", tap on a PDF page and on the area outside the page. For both taps, the app should show/hide navigation bar.

[ios-provider-pdf-renderer PR#4](https://github.com/ThePalaceProject/ios-provider-pdf-renderer/pull/4)

Yes

Yes, documentation in code

@vladimirfedorov


mauricecarrier7 (approved):
Sorry, my PR should have merged after this one, I bumped the version to 66 to account for that, so you will need to bump to 67.

PR #183, opened 2022-08-08
Present Audiobook download alerts from main thread

mauricecarrier7:
**What's this do?**
Ensures alerts are presented from the main thread

https://www.notion.so/lyrasis/iOS-The-app-crashes-after-choosing-the-not-downloaded-chapter-in-TOC-in-Biblioboard-and-Palace-Mark-2cdb0d6ca9f84127bf64e8a048369ac8

Download audiobook, attempt to open chapter that has not been downloaded. App should not crash.

https://github.com/ThePalaceProject/ios-audiobooktoolkit/pull/22/files

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me, but please change the default name

PR #198, opened 2022-09-27
Fixes the need to download books on My Books tab

vladimirfedorov:
**What's this do?**
Fixes the need to re-download books when the user switches to My Books tab right after the app starts.

Books "disappear" from the "My Books" tab when this tab is open when the app starts. That happens due to race conditions — the app reads the book registry when the "Catalog" tab is active, but the "My Books" tab also synchronises identifiers with this registry. If it manages to load *before* "My Books" does sync, everything looks fine, otherwise the "My Books" tab considers all the books missing (the registry is empty) ([Ticket](https://www.notion.so/lyrasis/iOS-Materials-need-to-be-re-downloaded-after-closing-and-reopening-the-app-for-Mississippi-Library--895b69d7d1834964b65702f3dfef1d9e))

Kindly follow the steps in the ticket.

No

Yes

Code documentation.

@vladimirfedorov 

https://user-images.githubusercontent.com/941531/192881076-26196194-4774-49e2-b9ca-b5a77f2ad054.mov


mauricecarrier7 (approved):
LGTM

PR #202, opened 2022-09-30
Implement obfuscation to protect Transifex credentials

mauricecarrier7:
**What's this do?**
Integrates obfuscator to encode/decode transifex credentials

https://www.notion.so/lyrasis/Transifex-Integrate-and-Obfuscate-APIToken-and-Secret-8457dda7083a4b6192d63ed4c1b78d37?d=3402ac26a1c14be98922d7da94d90565#501ebb27dcf24047bfc51a1ab964ee39

This is unit tested

N/A

No

N/A
@mauricecarrier7


Palace/Security/Obfuscator.swift
```diff
@@ -0,0 +1,78 @@
...
+        let cipher = [UInt8](self.salt.utf8)
+        let length = cipher.count
+        
+        var encrypted = [UInt8]()
+        
+        for t in text.enumerated() {
+            encrypted.append(t.element ^ cipher[t.offset % length])
+        }
+        
+        #if DEVELOPMENT
```
vladimirfedorov, line 47: Where is `DEVELOPMENT` defined? I couldn't find this constant in the project, but we have `DEBUG` for Debug mode

Palace/Utilities/Localization/TransifexManager.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Palace
+//
+//  Created by Maurice Carrier on 9/30/22.
+//  Copyright © 2022 The Palace Project. All rights reserved.
+//
+
+import Foundation
+
+struct TransifexManager {
+  static let token: [UInt8] = [101, 127, 97, 112, 66, 71, 38, 7, 15, 7, 80, 2, 76, 92, 125, 103, 42, 86,
```
vladimirfedorov, line 12: Is it ok that we keep token here and not in secrets?

Palace/Security/Obfuscator.swift
mauricecarrier7: This is borrowed code, this should be removed.

Palace/Utilities/Localization/TransifexManager.swift
mauricecarrier7: Ah, completely missed TPPSecrets, where is the encoding function, I can reuse the salt to encode the token

mauricecarrier7: Ah, I recreated the existing obfuscator. Will update

PR #208, opened 2022-10-11
Integrate Transifex SDK

mauricecarrier7:
**What's this do?**
-  Integrates TransifexSDK 
- Cleans up untranslated strings

https://www.notion.so/lyrasis/Transifex-Integrate-and-create-string-upload-scripts-63b2ec89a63e43698572ea2a738346eb

Change local language on device

N/A

N/A

N/A

@mauricecarrier7 

https://user-images.githubusercontent.com/6921353/195191540-5fe7f9d9-f94a-41a9-85b8-14ef5806aed2.mov


vladimirfedorov (approved):
Looks good do me.

Do I understand this right: now, when we need to change the name of a button, we don't change it in the app, but in Transifex instead? To keep the same localized string key we added to Transifex?

mauricecarrier7:
> Looks good do me.
> 
> Do I understand this right: now, when we need to change the name of a button, we don't change it in the app, but in Transifex instead? To keep the same localized string key we added to Transifex?

Thats correct., the SDK will fetch the most updated translations on each startup when the cache is invalid.  Unless we are adding or removing strings, we only need to update the Transifex web portal to change copy. When new strings are added or removed from the app itself, we will need to use the CLI tool to update the source strings.

PR #212, opened 2022-10-18
Bump build number

mauricecarrier7:
**What's this do?**
Bumps build number


vladimirfedorov (changes requested):
This target is `Palace-noDRM`, it alone won't start the build and upload process, `Palace` build number should also be changed.

vladimirfedorov (approved):
Looks good to me!

PR #215, opened 2022-11-10
Refactor book registry and related code

vladimirfedorov:
**What's this do?**
This fixes a lot of issues we previously had with TPPBookRegistry. The code is in Swift now, and the book registry doesn't expose methods that can potentially break its internal consistency.

Originally the problem was described in [this ticket](https://www.notion.so/lyrasis/iOS-Materials-need-to-be-re-downloaded-after-closing-and-reopening-the-app-for-Mississippi-Library--895b69d7d1834964b65702f3dfef1d9e). After several fixes the need to refactor the registry became obvious.

"My Books" section should keep books, bookmarks, reading positions, audiobook playhead positions in all cases described in the ticket.

NA

Yes

Yes, moved code documentation from the previous version and added some new.

@vladimirfedorov 

https://user-images.githubusercontent.com/941531/201204729-cbfb5b4e-af25-4132-8712-6baba20c351e.mov


mauricecarrier7 (approved):
This works great and is a huge improvement to the code base. LGTM

PR #216, opened 2022-11-17
Fix catalog books button state

vladimirfedorov:
**What's this do?**
 Fixes some bugs added along with the new book registry:
- Fixes the state of books in the Catalog (this also fixes the "cancelled" error after a book finishes downloading),
- Adds missing completion handler for background fetches.

There is the “Download” button instead of “Get” button in the Catalog [ticket](https://www.notion.so/lyrasis/iOS-There-is-Download-button-instead-of-Get-in-the-Catalog-4e881a142dc5470f871ff38e766c988b).

Please follow the steps in the [ticket](https://www.notion.so/lyrasis/iOS-There-is-Download-button-instead-of-Get-in-the-Catalog-4e881a142dc5470f871ff38e766c988b).

No

Yes

Code documentation

@vladimirfedorov 

https://user-images.githubusercontent.com/941531/202545210-7800cc50-34cd-4733-94c7-bc9736b2e0ac.mov


mauricecarrier7 (approved):
LGTM!

PR #217, opened 2022-11-18
Fix build errors in Palace-noDRM target

vladimirfedorov:
**What's this do?**
- Adds missing files to Palace-noDRM build
- Adds a step to GitHub workflow to verify that Palace-noDRM target builds without errors
- Fixes a crash when nil button is added

No ticket, just found we can't build Palace-noDRM anymore, which  makes testing without DRM impossible.

Select the `Palace-noDRM` target and run the app. The app should run without an issue.

No

No

NA

@vladimirfedorov

### DIFF
--- a/.github/workflows/unit-testing.yml
+++ b/.github/workflows/unit-testing.yml
@@ -31,6 +31,10 @@ jobs:
         run: ./scripts/build-3rd-party-dependencies.sh
         env:
           BUILD_CONTEXT: ci
+      - name: Build Palace without DRM support
+        run: ./scripts/xcode-build-nodrm.sh
+        env:
+          BUILD_CONTEXT: ci
       - name: Run Palace unit tests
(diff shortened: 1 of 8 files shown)


vladimirfedorov:
Found this while trying to build in Xcode 14 without ReadiumLCP framework — it's possible we'll need to request an updated version of the binary.

mauricecarrier7:
@vladimirfedorov , nice catch. Should we have a build phase to test the non-DRM target?

vladimirfedorov:
> @vladimirfedorov , nice catch. Should we have a build phase to test the non-DRM target on CircleCI?

I was trying to avoid that because it takes so much time, but probably it's better to have this step

vladimirfedorov:
Added a CI step to verify that Palace-noDRM can be built

mauricecarrier7 (approved):

PR #229, opened 2022-12-12
Reset initial system style when exiting an epub

mauricecarrier7:
**What's this do?**
Resets initial system style when exiting epub

https://www.notion.so/lyrasis/iOS-The-general-device-information-such-as-battery-Internet-connection-time-is-hidden-after-cho-b7f6ae7354ed4f3785cf38f54b833c07

 - Enter an epub in dark/light mode
 - change style of epub to dark mode
 - exit epub, app should be in initial mode

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me

PR #247, opened 2023-01-30
Fix outstanding localization issues

mauricecarrier7:
**What's this do?**
 This adds those translations to Localization files so the system will translate the back button on SwiftUI views.
Also adds translations for location tracking message

The system does not appear to recognize Transifex translations for the system back button. This adds those translations to Localization files so the system will translate the back button on SwiftUI views.

View TPPSettings view and child view, observe back button

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me

PR #258, opened 2023-03-02
Reuse object to reduce memory pressure

mauricecarrier7:
**What's this do?**
Updates local content deletion logic to use existing data object in order to reduce memory pressure that was causing the app to crash on large books. 

https://www.notion.so/lyrasis/Returning-Anna-Karenina-audiobook-causes-crash-on-iOS-6f0da86f0c0944b4b5297ddc5ced9428

Download one of the titles listed in the ticket and attempt to return/delete it. The app should not crash.

N/A

Yes

N/A

@mauricecarrier7

### DIFF
--- a/Palace/MyBooks/TPPMyBooksDownloadCenter.m
+++ b/Palace/MyBooks/TPPMyBooksDownloadCenter.m
@@ -555,7 +555,8 @@ didCompleteWithError:(NSError *)error
   if (!data) {
     return;
   }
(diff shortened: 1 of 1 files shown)


mauricecarrier7:
No need to bump the build version, it can go in before the other to outstanding PRs and be tested in the same build.

vladimirfedorov (approved):
Looks good to me

PR #260, opened 2023-03-21
Implement more selective remote position caching for audiobooks

mauricecarrier7:
**What's this do?**
- Throttles listening position save calls to server, saves once every 15 seconds while playing
- saves location on play, pause, seek, skip and chapter selection
- resolves TOC audiobook location server syncing issue

https://www.notion.so/lyrasis/iOS-Save-listening-position-less-frequently-6445ef4deb424103b2ec43e77541c418

- Play an audiobook, monitor console to evaluate time between location saves
- press play, notice save operation
- press pause, notice save operation
- seek, notice save operation
- scrub, notice save operation
- select chapter from spine, notice save operation

https://github.com/ThePalaceProject/ios-audiobooktoolkit/pull/34/files

Yes

N/A

@mauricecarrier7


Palace/Book/UI/TPPBookCellDelegate.m
```diff
@@ -40,6 +39,8 @@ @interface TPPBookCellDelegate () <RefreshDelegate>
 
 @implementation TPPBookCellDelegate
 
+static int ServerUpdateDelay = 15;
```
vladimirfedorov, line 42: It's a variable, would be better to start with a lower case s

Palace/Book/UI/TPPBookCellDelegate.m
```diff
@@ -500,8 +503,11 @@ - (void)pollAudiobookReadingLocation
      setLocation:[[TPPBookLocation alloc] initWithLocationString:string renderer:@"NYPLAudiobookToolkit"]
      forIdentifier:self.book.identifier];
     
-    // Save updated location on server
-    [TPPAnnotations postListeningPositionForBook:self.book.identifier selectorValue:string];
+    if ([[NSDate date] timeIntervalSinceDate: self.lastServerUpdate] >= ServerUpdateDelay) {
```
vladimirfedorov, line 506: Does it mean that in worst case scenario the app doesn't save last 15 seconds of progress? I couldn't find if we upload listening position if the user closes the book.

vladimirfedorov (changes requested):

Palace/Book/UI/TPPBookCellDelegate.m
vladimirfedorov: I think saving on exit when the user taps the Back button should be enough

mauricecarrier7: Resolved here: https://github.com/ThePalaceProject/ios-audiobooktoolkit/pull/35

PR #263, opened 2023-03-23
Fix/fetch library logos from s3

mauricecarrier7:
**What's this do?**
Updates account image loader to use S3 image instead of fetching auth docs for each account

Reduce server traffic

Load app, monitor networking traffic, auth docs should not be requested on startup

N/A

Yes

N/A

@mauricecarrier7


Palace/MyBooks/MyBooks/MyBooksView.swift
```diff
@@ -19,7 +19,8 @@ struct MyBooksView: View {
   var body: some View {
     NavigationLink(destination: accountScreen, isActive: $model.showAccountScreen) {}
     NavigationLink(destination: searchView, isActive: $model.showSearchSheet) {}
-    
+    NavigationLink(destination: accountPickerList, isActive: $showLibraryAccountView) {}
```
mauricecarrier7, line 22: I noticed a bug where when the user selected "add account" from the picker view the account list was not being presented

vladimirfedorov (approved):
Looks good to me!

PR #266, opened 2023-03-29
Set the correct source when posting annotations

mauricecarrier7:
**What's this do?**
Sets the correct source when posting annotations

https://www.notion.so/lyrasis/iOS-Save-listening-position-less-frequently-6445ef4deb424103b2ec43e77541c418?d=07e96cf92d874b889fcd516be8c758d6

Load an audiobook, press play, press pause or skip to set an annotation. Monitor network traffic and ensure that the book urn is passed as the Source parameter.

(https://github.com/ThePalaceProject/ios-audiobooktoolkit/pull/37)

yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me

PR #268, opened 2023-04-06
Set lat/long when creating account

mauricecarrier7:
**What's this do?**
- Resolves issue where account creation url was not loading in iOS 16.4
- Sends longitude and latitude coordinates in url when creating an account

https://www.notion.so/lyrasis/VLC-Web-form-does-not-load-in-app-to-create-virtual-library-card-on-iOS-1265878aafc742dc9b7b5d6d58178a33
https://www.notion.so/lyrasis/iOS-Pass-lat-and-long-as-URL-parameters-to-card-creator-4e28bfb8d4e74800bbb01bc3f47ee590

- Go to a library that offers virtual library cards: Morrill 
- Attempt to create a library card
- Application should request location permissions once and load webview successfully once permissions are granted

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me!

PR #275, opened 2023-04-20
Update reservation message copy

mauricecarrier7:
**What's this do?**
Updates reservation message copy to only show place in line when there are 0 books available

We are currently showing 0 books available

View reservation message for a title with 0 available copies

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me!

PR #286, opened 2023-05-22
Fix/save audiobookmarks locally

mauricecarrier7:
**What's this do?**
This implements local bookmarks and resolves several bugs.

https://www.notion.so/lyrasis/Bookmarks-are-not-added-for-some-Palace-Marketplace-audiobooks-e8528c31779c49ca93f502b49a833079
https://www.notion.so/lyrasis/iOS-The-deleted-bookmarks-briefly-reappear-in-the-Bookmark-tab-8f8816b34b354023b84f6c54d484a67e
https://www.notion.so/lyrasis/Save-audiobook-bookmarks-locally-f311f23c62b44eecab8234a094bebe7f

Disable sync
- Add and remove bookmarks
Enable sync
- Add and remove bookmarks

https://github.com/ThePalaceProject/ios-audiobooktoolkit/pull/42

Yes

N/A

@mauricecarrier7

### DIFF
--- a/Palace/Book/UI/AudiobookBookmarkBusinessLogic.swift
+++ b/Palace/Book/UI/AudiobookBookmarkBusinessLogic.swift
@@ -1,0 +1,190 @@
+//
+//  AudiobookBookmarkBusinessLogic.swift
+//  Palace
+//
+//  Created by Maurice Carrier on 4/12/23.
+//  Copyright © 2023 The Palace Project. All rights reserved.
+//
+
+import Foundation
+import NYPLAudiobookToolkit
+
+@objcMembers public class AudiobookBookmarkBusinessLogic: NSObject, AudiobookBookmarkDelegate {
+  var book: TPPBook
+  private var isSyncing: Bool = false
+  
+  init(book: TPPBook) {
+    self.book = book
+  }
+
+  public func saveListeningPosition(at location: String, completion: ((_ serverID: String?) -> Void)? = nil) {
+    TPPAnnotations.postListeningPosition(forBook: self.book.identifier, selectorValue: location, completion: completion)
+  }
+
+  // Need to replace exiting bookmark if it already exists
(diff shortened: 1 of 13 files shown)


Palace/Book/UI/AudiobookBookmarkBusinessLogic.swift
```diff
@@ -0,0 +1,230 @@
...
+  
+  private func replace(oldLocation: ChapterLocation, with newLocation: ChapterLocation) {
+    guard
+      let oldLocation = oldLocation.toTPPBookLocation(),
+      let newLocation = newLocation.toTPPBookLocation() else { return }
+    registry.replaceGenericBookmark(oldLocation, with: newLocation, forIdentifier: book.identifier)
+  }
+}
+
+extension Array where Element == ChapterLocation {
```
vladimirfedorov, line 135: Maybe nitpicking, can we move this array extension and the ChapterLocation extension below to a separate file if they shouldn't be in NYPLAudiobookToolkit? Also would be great to add another ChapterLocation extension from TPPBookRegistry+Extensions.swift to that file as well )

vladimirfedorov (approved):
I have a code organization suggestion, but otherwise looks good to me.

Palace/Book/UI/AudiobookBookmarkBusinessLogic.swift
mauricecarrier7: I agree with moving the ChapterLocationExtension to its own file. But I think that this should stay here because it is a very specific implementation that should only be used in the context that this class uses it.

mauricecarrier7: @vladimirfedorov  Updates here: https://github.com/ThePalaceProject/ios-audiobooktoolkit/pull/44

mauricecarrier7:
The test run with no issues individually but appear to be flaky when run together, likely due to state corruption. Will disable for now.

PR #288, opened 2023-05-26
Resolves missing alert at the end of Overdrive and Biblioboard audiobooks

mauricecarrier7:
**What's this do?**
Ensure that audiobook return alert is displayed at the end of Overdrive and Biblioboard books

https://www.notion.so/lyrasis/iOS-There-is-no-alert-Your-audiobook-has-finished-Would-you-like-to-return-it-at-the-end-of-Bi-39e52600db0247c1ba4af8d79574bc56

- Download an Overdrive or Biblioboard audiobook
- Play to the end of the book
- The alert should be displayed 

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me!

PR #290, opened 2023-06-02
Fix deep linking on first run

vladimirfedorov:
**What's this do?**
- Adds missing sign in screen on first run when current account is not the account the user is trying to sign in with;
- Removes welcome screen if it is present

Review comments to [this ticket](https://www.notion.so/lyrasis/iOS-Implement-deep-linking-for-library-sign-in-920653e4c9524e4c9e47a110bca66fc3).

- Terminate the app,
- Open a URL containing a library link on an iOS device (e.g., one from the ticket), the link should take you to the library login screen and automatically paste a barcode into the form.

NA

Yes

Yes, code documentation

@vladimirfedorov


mauricecarrier7 (approved):
LGTM

PR #307, opened 2023-08-08
Delete cache on app uninstall

mauricecarrier7:
**What's this do?**
Deletes keychain when a new install occurs

https://ebce-lyrasis.atlassian.net/browse/PP-239

Install app, login to account, delete app, reinstall, download the same library, login info should be empty.

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me

PR #337, opened 2023-10-02
[PP-467] Token expiration message displays to user

mauricecarrier7:
**What's this do?**
Ensures token refresh takes place automatically before errors are displayed to the user.

https://ebce-lyrasis.atlassian.net/browse/PP-467
https://ebce-lyrasis.atlassian.net/browse/PP-519

Log into A1QA or Lyrasis Reads and download a title. Wait 1 hour and download another. Download should progress as expected and now error should be shown. 

N/A

Yes

N/A

@mauricecarrier7


Palace/Network/TPPNetworkExecutor.swift
```diff
@@ -78,11 +78,33 @@ extension TPPNetworkExecutor: TPPRequestExecuting {
...
+    refreshTokenAndResume(task: nil) { [weak self] newToken in
+      guard let strongSelf = self else { return }
+      
+      if let token = newToken {
+        var updatedRequest = req
+        updatedRequest.setValue("Bearer \(token)", forHTTPHeaderField: "Authorization")
+        strongSelf.executeRequest(updatedRequest, completion: completion)
+
+      } else {
+        completion(NYPLResult.failure(NSError(domain: "com.PalaceApp.error", code: 401, userInfo: [NSLocalizedDescriptionKey: "Unauthorized HTTP"]), nil))
```
vladimirfedorov, line 96: There's TPPErrorLogger with error codes, and code `401` means audiobook is corrupted ([here](https://github.com/ThePalaceProject/ios-core/blob/main/Palace/Logging/TPPErrorLogger.swift)). You may need to add a new code there.

Also, we have `TPPErrorLogger.clientDomain` for NSError domains to not type domain values everywhere, but I think it was used once or twice so far.

mauricecarrier7: Ah, thanks, was looking for this.

vladimirfedorov (approved):
Looks good to me

PR #339, opened 2023-10-03
[PP-522] Loan limit message not shown

mauricecarrier7:
**What's this do?**
Ensure proper error handling so loan limit message is presented
Also only refresh token before performing task if the request is token based

https://ebce-lyrasis.atlassian.net/browse/PP-522

Attempt to download more than the allowed number of titles from a given library, error message should be presented.

N/A

Yes

N/A

@mauricecarrier7


vladimirfedorov (approved):
Looks good to me

PR #342, opened 2023-10-13
[PP-452] Implement Epub search

mauricecarrier7:
**What's this do?**
- Adds search functionality to Epub

https://ebce-lyrasis.atlassian.net/browse/PP-425

- Download an epub
- open book
- select search window
- enter a search string and scroll to see list dynamically update

N/A

Yes

N/A

@mauricecarrier7


Palace/Reader2/UI/TPPEPUBViewController.swift
```diff
@@ -197,3 +208,22 @@ extension TPPEPUBViewController: UIPopoverPresentationControllerDelegate {
...
+    }
+  
+    if let navigator = navigator as? DecorableNavigator {
+      
+      var decorations: [Decoration] = []
+      decorations.append(Decoration(
+        id: "tts-search",
+        locator: location,
+        style: .highlight(tint: .red)))
+      navigator.apply(decorations: decorations, in: "tts")
```
vladimirfedorov, line 226: may be we should use `search` here, as `tts` means text-to-speach

mauricecarrier7: ah I see, thanks, will update

vladimirfedorov:
Github workflows occasionally fail with "Could not find any available simulators for iOS" error since Friday, if you see it try to restart the job, it sometimes helps.

PR #369, opened 2023-11-30
[PP-694] Use token auth for annotations

mauricecarrier7:
**What's this do?**
Uses TPPNetworkExecutor for all Annotations networking to leverage existing token management pathways

https://ebce-lyrasis.atlassian.net/browse/PP-694

Use a network proxy agent such as Proxyman and observe networking calls while making annotations updates, add, delete etc. in Minotaur. Notice headers will include auth token.

N/A

Yes

N/A

@mauricecarrier7


Palace/Samples/AudiobookSamplePlayer/AudiobookSamplePlayer.swift
```diff
@@ -98,16 +98,16 @@ class AudiobookSamplePlayer: NSObject, ObservableObject {
   private func downloadFile() {
     state = .loading
 
-    let _ = sample.fetchSample { [unowned self]  result in
```
mauricecarrier7, line 101: This caused a crash during testing, so I resolved it here.

PR #506, opened 2025-01-30
Update logo initialization order for faster loading

jonathangreen:
**What's this do?**

I've wondered for awhile why the app seems so slow to load the main catalog screen sometimes. Looking at the requests, it seems like its because we are adding all the library logos to the network queue, before adding the libraries authentication document. 

When the cache is stale this means that we need to fetch all ~600 library logos before displaying the main catalog view. I addressed this by not adding the task to get the library logo when the `Account` class is initialized, but providing a method to do so, then calling this method after the libraries authentication document is added to the queue.

I'm not sure this is the cleanest approach, and I don't really understand the order things are expected to be initialized in here, so feel free to give this a tough code review. It does help the user experience a lot when the cache is empty, I see a 10 - 15 second reduction in the time it takes to display the main catalog page.

Honestly because the long load time has been driving me crazy for some time.

[Description of any tests that need to be performed once merged goes here]

[Description of any watchouts, dependencies, or issues we should be aware of goes here]

[Bump the Palace build number to generate a new build on ThePalaceProject/ios-binaries]

### DIFF
--- a/Palace/Accounts/Library/Account.swift
+++ b/Palace/Accounts/Library/Account.swift
@@ -397,7 +397,8 @@ protocol AccountLogoDelegate: AnyObject {
       details = AccountDetails(authenticationDocument: authenticationDocument, uuid: uuid)
     }
   }
-  
+  var logoUrl: URL? = nil
+
 
   var loansUrl: URL? {
     return details?.loansUrl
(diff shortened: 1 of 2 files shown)


mauricecarrier7 (approved):
AccountManager could use a ton of improvement, but this is a great start. I don't see immediate issues here. LGTM!

PR #119, opened 2022-01-28
Update bookmarks to R2 spec and resolve bookmark syncing issue

mauricecarrier7:
**What's this do?**
Update bookmarks to R2 spec and resolve bookmark syncing issue

**Why are we doing this? (w/ Notion link if applicable)**
[Notion](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=490d677e84b34ca9935a225ef0ac120e&p=cfcc3d906c6f451483602c7fb41e1974)

**How should this be tested? / Do these changes have associated tests?**
Set bookmarks on one device, attempt to access the same locations on another device.

**Dependencies for merging? Releasing to production?**
N/A

**Does this include changes that require a new Palace build for QA?**
Yes

**Has the application documentation been updated for these changes?**
N/A

**Did someone actually run this code to verify it works?**
@mauricecarrier7

PR #509, opened 2025-02-03
Make sure binaries are deleted when returing an book (PP-1024)

jonathangreen:
**What's this do?**

Updates `MyBooksDownloadCenter` so that both books and audiobooks are correctly deleted from the device when the item is returned.

See: PP-1024

Checking out various content sources and make sure the item is deleted.

This relies on the PR to the audiobook toolkit: 
https://github.com/ThePalaceProject/ios-audiobooktoolkit/pull/119

[Bump the Palace build number to generate a new build on ThePalaceProject/ios-binaries]

I tested it against:
Audiobooks:
- Findaway
- Open access
- Marketplace LCP
- Marketplace non-LCP
Books:
- Adobe

And I saw the space being released in all those cases.

### DIFF
--- a/Palace/MyBooks/MyBooksDownloadCenter.swift
+++ b/Palace/MyBooks/MyBooksDownloadCenter.swift
@@ -435,9 +435,10 @@ import OverdriveProcessor
 
 extension MyBooksDownloadCenter {
   func deleteLocalContent(for identifier: String, account: String? = nil) {
+    let current_account: String? = account ?? AccountsManager.shared.currentAccountId
     guard let book = bookRegistry.book(forIdentifier: identifier),
-          let bookURL = fileUrl(for: identifier, account: account) else {
-      NSLog("WARNING: Could not find book to delete local content.")
+          let bookURL = fileUrl(for: identifier, account: current_account) else {
+      Log.warn(#file, "Could not find book to delete local content \(identifier)")
       return
     }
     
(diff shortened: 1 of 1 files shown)


Palace/MyBooks/MyBooksDownloadCenter.swift
```diff
@@ -435,9 +435,10 @@ import OverdriveProcessor
 
 extension MyBooksDownloadCenter {
   func deleteLocalContent(for identifier: String, account: String? = nil) {
+    let current_account: String? = account ?? AccountsManager.shared.currentAccountId
     guard let book = bookRegistry.book(forIdentifier: identifier),
-          let bookURL = fileUrl(for: identifier, account: account) else {
-      NSLog("WARNING: Could not find book to delete local content.")
+          let bookURL = fileUrl(for: identifier, account: current_account) else {
+      Log.warn(#file, "Could not find book to delete local content \(identifier)")
```
jonathangreen, line 441: Most of the code in this file is logging via `NSLog`, but it looks like most of the Palace code is using `Log`, so I wasn't sure which way to go for log statements here. 

Let me know what you prefer for new code @mauricecarrier7. I don't mind switching this up if its something other then `Log`.

Palace/MyBooks/MyBooksDownloadCenter.swift
```diff
@@ -435,9 +435,10 @@ import OverdriveProcessor
 
 extension MyBooksDownloadCenter {
   func deleteLocalContent(for identifier: String, account: String? = nil) {
+    let current_account: String? = account ?? AccountsManager.shared.currentAccountId
```
jonathangreen, line 438: This is the main fix here, we were getting `nil` for `account` when deleting content, so we were unable to find the binaries.

Palace/MyBooks/MyBooksDownloadCenter.swift
```diff
@@ -451,69 +452,32 @@ extension MyBooksDownloadCenter {
...
-    audiobook.deleteLocalContent(completion: { _, _ in })
-    if FileManager.default.fileExists(atPath: bookURL.path) {
-      try? FileManager.default.removeItem(at: bookURL)
+
+    // LCP Audiobooks are a single binary file, without an easily loaded manifest. So they skip this logic
+    // that deleted the local audio files, used by other audiobook types.
+    if (!isLcpAudiobook) {
+      let manifestData = try Data(contentsOf: bookURL)
+      let manifest = try Manifest.customDecoder().decode(Manifest.self, from: manifestData)
+      AudiobookFactory.audiobookClass(for: manifest).deleteLocalContent(manifest: manifest, bookIdentifier: book.identifier)
```
jonathangreen, line 478: This is very similar to `deleteAudiobookContent` function that we previously had, however once the `nil` account issue was fixed, it exposed that creating an `Audiobook` object was creating a player, this caused the audiobook to start playing, which then caused `deleteLocalContent` to fail. So I moved `deleteLocalContent` out to a class method, so it can be called without creating an `Audiobook` first.

Palace/MyBooks/MyBooksDownloadCenter.swift
mauricecarrier7: Log is fine, that was likely just overlooked

PR #524, opened 2025-02-27
Fix unsupported OD holds appearing in my books (PP-2154)

jonathangreen:
**What's this do?**

If you take out a hold on Overdrive content that isn't supported in the Libby app, its showing up in Palace as downloadable. This updates that logic to make sure that it doesn't end up in that state.

Note: I tried to sort out exactly why the code I pulled out was here, but I don't understand what we were checking for with the `actuallyOnHold` flag, which makes me a bit nervous. Let me know if this looks like it will cause any problems to you.

See JIRA: [PP-2154](https://ebce-lyrasis.atlassian.net/browse/PP-2154)

I've got a barcode where this is currently happening I can share. Holds that are currently showing up in the my books screen should not be shown at all.

None.

Might be good to have this go out in a new build so Courtney / Carissa can verify it.

I did against the barcode i mentioned earlier.

[PP-2154]: https://ebce-lyrasis.atlassian.net/browse/PP-2154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Palace/Book/Models/TPPBookRegistryRecord.swift
+++ b/Palace/Book/Models/TPPBookRegistryRecord.swift
@@ -28,8 +28,7 @@ class TPPBookRegistryRecord: NSObject {
     self.genericBookmarks = genericBookmarks
     
     super.init()
-    
-    var actuallyOnHold = false
+
     if let defaultAcquisition = book.defaultAcquisition {
       defaultAcquisition.availability.matchUnavailable { _ in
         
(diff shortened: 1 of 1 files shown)


mauricecarrier7 (approved):
I'm not sure what impact removing actuallyOnHold will have, but it looks like it was initially added to resolve a book button state issue. I'll keep this in mind if we see any potentially related issues arise. 

Seems like there is a case where the Record is initialized with a holding state that doesn't match it's acquisition state and we are forced resetting it. But seems like a bandaid. If anything removing this logic will surface the root cause.

PR #824, opened 2026-04-10
Fix pinless login (PP-4045), hold cancel (PP-4033), error origin tagging

mauricecarrier7:
## Summary

Three fixes cherry-picked from release/2.2.5 into modernize/whole-shot:

**PP-4045 (BLOCKER):** Libraries that don't require a PIN (e.g. Wolcott Public Library, Bentley Memorial Library) got "Invalid Credentials" on sign-in. The `TokenRequest` guard added in March (`77c5af2e5`) rejected empty passwords, but empty password is valid Basic Auth for pinless libraries per RFC 7617. Fix: only guard against empty username.

**PP-4033:** Cancel Hold button did nothing. The revoke endpoint error handler only handled `TypeNoActiveLoan` and `TypeInvalidCredentials`. Any other server error type fell through silently with no user feedback. Fix: catch-all else branch that calls `announceReturnFailed()`.

**Error origin tagging (F-META-1):** Every non-fatal error logged to Crashlytics now includes `error_origin` (client/server/network/vendor) as a custom key. Auto-classified from error codes, HTTP status codes, and URL error domains. Enables filtering the Crashlytics dashboard to separate client bugs from server/network/vendor issues.

## Test plan

- [x] 13 credential guard tests pass (0 failures)
- [x] New pinless login test verifies Basic Auth "barcode:" format with empty password
- [x] Build succeeds on Palace scheme
- [ ] Manual: sign into Wolcott Public Library or Bentley Memorial Library with barcode only (no PIN)
- [ ] Manual: place and cancel a hold on A1QA Test Library

### DIFF
--- a/Palace/AppInfrastructure/FirebaseManager.swift
+++ b/Palace/AppInfrastructure/FirebaseManager.swift
@@ -283,6 +283,13 @@ final class FirebaseManager {
         #endif
     }
 
+    /// Sets a custom key-value pair on Crashlytics for dashboard filtering.
+    func setCrashlyticsKey(_ key: String, value: String) {
+        #if FEATURE_CRASH_REPORTING
+        Crashlytics.crashlytics().setCustomValue(value, forKey: key)
+        #endif
+    }
+
     // MARK: - Analytics Events
 
     /// Logs an enhanced error event to Analytics.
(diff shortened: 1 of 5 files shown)


mauricecarrier7:
Changes already included in #822 (per-account credentials). Closing as merged.

PR #832, opened 2026-04-15
Fix empty device ID breaking cross-device bookmark sync

mauricecarrier7:
## Summary

- Non-Adobe-DRM users (LCP, Findaway, open-access) had annotation device field set to empty string, breaking cross-device sync detection
- The "sync to furthest position?" prompt never triggered for these users because both devices appeared identical
- Adds `AnnotationDevice.currentID()` which falls back to `FirebaseManager.shared.deviceID` when Adobe DRM ID is unavailable

## How this was found

During PP-4020 regression testing, we ran an automated API-level sync test against the A1QA staging server (`gorgon.staging.palaceproject.io/a1qa-test/annotations/`). The test script (`scripts/test-sync.sh`) queried all 255 annotations on the account and grouped them by device ID. This revealed:

- **23 annotations with empty device IDs** (`device=""`) — all from non-DRM test sessions (April 2-13)
- **232 annotations with proper `urn:uuid:` device IDs** — from devices with Adobe DRM activated

Cross-referencing `TPPAnnotations.swift:134` confirmed the root cause: `TPPUserAccount.sharedAccount().deviceID ?? ""` — the `deviceID` property is only set during Adobe RMSDK activation, so LCP/Findaway/open-access users always get the empty fallback.

The empty device ID breaks the sync comparison in `TPPLastReadPositionSynchronizer.swift:85` — when `deviceID == drmDeviceID` and both are empty, the app treats every server annotation as "from this device" and never shows the sync prompt.

## Changes

| File | Change |
|------|--------|
| `TPPAnnotations.swift` | Add `AnnotationDevice.currentID()` — prefers Adobe DRM ID, falls back to `urn:uuid:{FirebaseManager.deviceID}` |
| `TPPAnnotations.swift` | Replace `TPPUserAccount.sharedAccount().deviceID ?? ""` with `AnnotationDevice.currentID()` (2 call sites) |
| `ReaderService.swift` | Use `AnnotationDevice.currentID()` for sync comparison (was `TPPUserAccount.deviceID`) |
| `TPPBaseReaderViewController.swift` | Use `AnnotationDevice.currentID()` for bookmark business logic (was `TPPUserAccount.deviceID`) |
| `TPPAnnotationsTests.swift` | 3 new tests: non-empty guarantee, stability, format validation |

## Test plan

- [x] `AnnotationDeviceIDTests` — 3 tests pass (non-empty, stable, urn:uuid: format)
- [ ] Manual: Open EPUB on Sim A → read to 40% → open same book on Sim B → verify "sync to furthest position?" prompt appears
- [ ] Manual: Create bookmark on Sim A → open same book on Sim B → verify bookmark appears in list
- [ ] Verify Adobe DRM users still get their existing device ID (backwards compatible)


mauricecarrier7:
Superseded by new PR based off modernize/whole-shot

PR #840, opened 2026-04-16
Use existing github action for doing jira release sync

dbernstein:
**Use existing github action for doing jira release sync**
---

**What's this do?**
Replaces the existing jira release sync workflow contents with a reference to the plugin in ThePalaceProject/circulation.

Trying to reduce code duplication.

It should just work next time we do a release.
N/A
No
N/A
The action is being used by other repositories (such circulation-admin)


dbernstein:
Closing - I was looking at the wrong branch.  I realized that this was already done and merged into develop.

PR #845, opened 2026-04-21
PP-4129: Orphaned-download auto-recovery + finish TPPBookRegistry facade

mauricecarrier7:
## Summary

Fixes [PP-4129](https://ebpc.atlassian.net/browse/PP-4129): patrons stuck in "weird state" where the registry says a book is downloaded but the file is missing on disk, causing infinite loops on open/delete (HelpSpot #17564, #17546, #17648, #17669).

Also completes the `TPPBookRegistry` facade decomposition that was started in `5a1ae2a42` but left half-done — the monolith continued to drive all mutations directly while `BookRegistryStore` / `BookRegistrySync` / `BookmarkManager` sat unused. This PR wires the facade so the decomposed types are actually the load-bearing implementation.

### Fixes

1. **Orphan auto-recovery (PP-4129):** during registry load, any book whose state is `downloadSuccessful` but whose file does not exist on disk is flipped back to `downloadNeeded` and auto-restarted (5s after load, only if the active account still matches).
2. **Account-capture on save:** mutations now capture `currentAccount.uuid` synchronously at dispatch time and thread it through to the save closure. Previously an async save queued during a library switch could land in the wrong account's registry file (e.g. writing A1QA records into the Icarus registry).
3. **UI race on account change:** the `.TPPCurrentAccountDidChange` observer synchronously clears `registrySubject` and sets `state = .loading` before kicking off `load()`, so a user who taps a book during the load window no longer interacts with the prior library's registry snapshot.
4. **Swift exclusivity (mutateRegistry onComplete):** `BookRegistryStore.mutateRegistry(_:onComplete:)` now dispatches `onComplete` as a **separate** barrier job after `block(&self.registry)` has fully completed, so `inout` exclusivity does not remain open when the completion handler reads `self.registry` (e.g. for snapshot-and-save). This resolves the "Simultaneous accesses to registry" launch crashes seen after the first landing attempt.

### Facade decomposition

- `TPPBookRegistry` is now a ~535-line thin facade that delegates to `BookRegistryStore` (thread-safe in-memory store), `BookRegistrySync` (load / save / server sync), and `BookmarkManager` (bookmark + location CRUD).
- All three components use **closure-based APIs** (no actors / async-await) to stay consistent with the rest of the codebase and the surrounding call sites.
- `BookmarkManager` methods take an explicit `account: String` captured by the caller, with a `didMutate` flag so `save(account)` only fires when the registry actually changed.

## Test plan

- [x] `BookRegistryStoreTests`, `BookRegistrySyncTests`, `BookmarkManagerTests`, `TPPBookRegistryTests`, `TPPBookRegistryLCPIntegrationTests` — **all pass, 0 failures** (98 tests in the registry core; 344 across the broader Book-domain suite).
- [x] Simulator QA (iPhone 16 Pro, iOS 18.4) end-to-end:
  - [x] Orphan recovery: delete a downloaded book's on-disk file, relaunch → book flips to `downloadNeeded` and auto-restarts after 5s.
  - [x] Account switch cross-contamination: download books in Icarus → switch to A1QA → switch back to Icarus → no stale Icarus state, no infinite loading, no cross-account writes.
  - [x] UI race during switch: tapping a book mid-load no longer lets you interact with the prior library's registry.
  - [x] Repeated launch / library switch cycles — no exclusivity crashes.

## Governance

ForgeOS changeset: `cs_858c0ff0` — all gates passed (review / testing / release).

[PP-4129]: https://ebce-lyrasis.atlassian.net/browse/PP-4129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release-on-merge.yml
+++ b/.github/workflows/release-on-merge.yml
@@ -19,6 +19,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          fetch-depth: 0
           token: ${{ secrets.CI_GITHUB_ACCESS_TOKEN }}
       - name: Checkout Certificates
         uses: actions/checkout@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          fetch-depth: 0
           token: ${{ secrets.CI_GITHUB_ACCESS_TOKEN }}
       - name: Checkout Certificates
         uses: actions/checkout@v3
--- a/.github/workflows/upload-on-merge.yml
+++ b/.github/workflows/upload-on-merge.yml
@@ -54,6 +54,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          fetch-depth: 0
           token: ${{ secrets.CI_GITHUB_ACCESS_TOKEN }}
 
       - name: Cache Swift Package Manager
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -91,6 +91,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          fetch-depth: 0
           token: ${{ secrets.CI_GITHUB_ACCESS_TOKEN }}
 
       - name: Cache Swift Package Manager
--- a/Palace/Book/Models/BookRegistryStore.swift
+++ b/Palace/Book/Models/BookRegistryStore.swift
@@ -65,12 +65,35 @@ class BookRegistryStore {
   }
 
   /// Provides direct write access to the registry dictionary within a barrier block.
-  func mutateRegistrySync(_ block: (_ registry: inout [String: TPPBookRegistryRecord]) -> Void) {
+  ///
+  /// `onComplete` runs on the barrier queue AFTER `block(&self.registry)` has
+  /// fully completed — it is dispatched as a separate barrier job so Swift's
+  /// exclusivity tracking does NOT consider the `inout` access still open during
+  /// `onComplete`. Use `onComplete` for snapshot-and-save work (reading the
+  /// registry) that cannot be done inside the mutation block itself without
+  /// triggering "Simultaneous accesses to registry, but modification requires
+  /// exclusive access" (PP-4129 crash fix).
+  func mutateRegistrySync(
+    _ block: (_ registry: inout [String: TPPBookRegistryRecord]) -> Void,
+    onComplete: (() -> Void)? = nil
+  ) {
     performBarrierSync { block(&self.registry) }
+    // Second barrier dispatch: inout access from `block` has fully ended before
+    // `onComplete` runs. onComplete can safely read `self.registry` (e.g. to
+    // snapshot it for save) without tripping exclusivity.
+    if let onComplete = onComplete {
+      performBarrierSync { onComplete() }
+    }
   }
 
-  func mutateRegistry(_ block: @escaping (_ registry: inout [String: TPPBookRegistryRecord]) -> Void) {
+  func mutateRegistry(
+    _ block: @escaping (_ registry: inout [String: TPPBookRegistryRecord]) -> Void,
+    onComplete: (() -> Void)? = nil
+  ) {
     performBarrier { block(&self.registry) }
+    if let onComplete = onComplete {
+      performBarrier { onComplete() }
+    }
   }
 
   // MARK: - Query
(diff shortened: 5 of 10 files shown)


mauricecarrier7:
## Self-review

GitHub doesn't let me approve my own PR, so posting findings as a comment.

### What's in the diff

**PP-4129 bug fix — orphaned downloads + account-cross-contamination (Palace/Book/Models/)**

- `BookRegistrySync.load()` now validates on-disk presence for every `downloadSuccessful` book, flipping stale records to `downloadNeeded` and scheduling a guarded 5s auto-restart. The guard re-checks `AccountsManager.shared.currentAccountId == loadedAccount` before firing, so switching libraries during the wait doesn't trigger a re-download under the wrong auth context.
- Every mutation path in `TPPBookRegistry` captures `accountsManager.currentAccount?.uuid` **synchronously at dispatch time** and passes it through `save(for:)`. Fixes the cross-contamination regression where an async save queued on account A could land in account B's registry file after a library switch.
- `TPPBookRegistry.setupAccountDidChangeObserver` synchronously sends `[:]` on `store.registrySubject` and flips `state = .loading` before kicking off `load()`, so UI readers can't briefly see the previous account's books during the load window.
- `BookRegistryStore.mutateRegistry(_:onComplete:)` dispatches `onComplete` as a **separate barrier job** after `block(&self.registry)` returns, so Swift's `inout` exclusivity doesn't remain open when the completion handler reads `self.registry` for snapshot-and-save. Resolves the "Simultaneous accesses to registry" launch crashes.

**Facade decomposition — completing commit `5a1ae2a42`**

- `TPPBookRegistry` shrinks from 987 lines of direct state management to ~535 lines of thin delegation. Matches the intended design where `BookRegistryStore` owns thread-safe storage, `BookRegistrySync` owns load/save/server sync, and `BookmarkManager` owns bookmark + location CRUD.
- All three collaborators use closure-based APIs (no actors/async-await) to stay consistent with surrounding call sites.
- `BookmarkManager` gets a `didMutate` flag so `save(account)` only fires when the mutation actually changed state — fixes two pre-existing "save-even-on-noop" test failures.

**CI — `fetch-depth: 0` on release-notes checkouts (.github/workflows/)**

- Added to `upload-on-merge.yml`, `upload.yml`, `release-on-merge.yml`, `release.yml`. Required by `mobile-certificates` PR #33 (`39fddc01`, 2026-04-08), which rewrote `ReleaseNotes.py` to walk local git log instead of the PR API. The April 21 upload failed because it was the first version-bumped develop merge since that rewrite, and the default `actions/checkout@v3` shallow clone had no tag refs for `git tag --merged HEAD` to find.
- Verified locally: script produces a clean 17-ticket v2 output against current HEAD with `2.2.5` as the baseline tag.

### Verification

- **98 tests** pass across `BookRegistryStore`, `BookRegistrySync`, `BookmarkManager`, `TPPBookRegistry`, and `TPPBookRegistryLCPIntegration`. 0 failures. 8 skipped (all pre-existing — see notes below).
- **344 tests** pass across the broader Book-domain suite.
- Simulator QA (iPhone 16 Pro, iOS 18.4):
  - Orphan recovery: delete a book's on-disk file, relaunch → state flips to `downloadNeeded`, auto-restarts 5s later ✅
  - Cross-account contamination: Icarus → A1QA → Icarus round trip, no stale state, no infinite loading ✅
  - UI race: tapping a book during library switch no longer reaches the prior registry ✅
  - Repeated library switches — no exclusivity crashes ✅

### Observations (non-blocking — follow-ups welcome)

1. **7 pre-existing skipped `BookRegistrySync` tests** were skipped before this PR. I tried unskipping one to see if the facade rewrite made them pass — they still fail because `AccountsManager.shared.currentAccountId` in the simulator shadows the test's isolated account, causing `load()` to load the simulator's real registry instead of the test fixture. Follow-up ticket candidate: `BookRegistrySync.load()` should treat a non-nil `account` parameter as authoritative and not fall back to the shared AccountsManager at all.
2. **Minor inconsistency:** `BookRegistrySync.load()` uses `AccountsManager.shared.currentAccountId`, while mutations in `TPPBookRegistry` use `accountsManager.currentAccount?.uuid`. These resolve to the same string; standardizing is cosmetic.
3. **`BoolWithDelay`** (pre-existing) doesn't cancel its reset timer in deinit. Minor leak risk only at registry teardown, which happens effectively never.

### ForgeOS governance

Changeset `cs_858c0ff0` — all gates promoted (review / testing / release). Release check: `can_release: true`.

PR #847, opened 2026-04-21
Filter OPDS2 publications with no in-app-renderable acquisition path

mauricecarrier7:
## Summary
- Palace Bookshelf books on the lyrasis-reads CM return as OPDS2 publications with one acquisition: `application/opds-publication+json` with a single indirect of `text/html;profile=http://librarysimplified.org/terms/profiles/streaming-media`. iOS has no in-app web reader, so these publications used to land in the catalog with no buttons and would crash into "format not supported" if the user managed to trigger `openBook` through any other path.
- After acquisition-link conversion, drop the publication when no acquisition resolves to a path supported by `TPPOPDSAcquisitionPath`. Both `OPDS2Publication.toBook()` and `OPDS2FullPublication.toBook()` get the filter so feed/search/details all stay consistent.

## Reproduction
Confirmed against the live staging feed at `https://gorgon.staging.palaceproject.io/lyrasis-reads/feed/2420?entrypoint=All` for `urn:uuid:84dac408-77ce-4afc-8393-9e0ced7ea3ef` ("CYTOMEGALOVIRUS INFECTION..."). The publication's only acquisition link:

```json
{
  "rel": "http://opds-spec.org/acquisition/borrow",
  "type": "application/opds-publication+json",
  "properties": {
    "indirectAcquisition": [
      { "type": "text/html;profile=http://librarysimplified.org/terms/profiles/streaming-media" }
    ]
  }
}
```

`text/html;profile=...streaming-media` is not in `supportedSubtypes[ContentTypeOPDSPublication]` (`Palace/OPDS/TPPOPDSAcquisitionPath.swift:102-115`). The acquisition path is empty, `defaultAcquisition` is nil, and the book is `.unsupported`. iOS Palace has never supported streaming-HTML books — Palace web/Android render them via web reader.

## Test plan
- [x] `testToBook_dropsPublicationWhenOnlyAcquisitionIsStreamingHTMLIndirect` — passes (0.005s). Exercises the exact CM-returned shape.
- [x] `testToBook_keepsPublicationWhenAtLeastOneAcquisitionIsSupported` — passes (0.004s). Mixed-format books are kept.
- [x] **Mutation test**: forcing `hasOpenablePath = true` makes the regression test fail (1.012s), reverting passes.
- [x] Full OPDS2 test suite (134 tests) passes after updating `testToBook_mapsSampleLink` to include a borrow link (sample-only publications are now correctly filtered — they would have been button-less anyway).
- [ ] Manual verification on lyrasis-reads staging: Palace Bookshelf swimlane should no longer show button-less books.

## Follow-up
A future PR could add `BookButtonType.readInBrowser` that hits the borrow URL and opens the resulting streaming HTML in `SFSafariViewController`, matching Palace web/Android behavior. That's intentionally out of scope here — this PR only stops the broken UI.

## ForgeOS
- Changeset: `cs_f2f12d94` / Initiative: `init_67d79332` (Stability Phase 3)
- Gates: review ✅ → testing ✅ → release ✅ (134 tests passed)

### DIFF
--- a/.github/workflows/release-on-merge.yml
+++ b/.github/workflows/release-on-merge.yml
@@ -19,6 +19,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          fetch-depth: 0
           token: ${{ secrets.CI_GITHUB_ACCESS_TOKEN }}
       - name: Checkout Certificates
         uses: actions/checkout@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          fetch-depth: 0
           token: ${{ secrets.CI_GITHUB_ACCESS_TOKEN }}
       - name: Checkout Certificates
         uses: actions/checkout@v3
--- a/.github/workflows/upload-on-merge.yml
+++ b/.github/workflows/upload-on-merge.yml
@@ -54,6 +54,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          fetch-depth: 0
           token: ${{ secrets.CI_GITHUB_ACCESS_TOKEN }}
 
       - name: Cache Swift Package Manager
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -91,6 +91,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          fetch-depth: 0
           token: ${{ secrets.CI_GITHUB_ACCESS_TOKEN }}
 
       - name: Cache Swift Package Manager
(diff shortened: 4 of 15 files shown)


mauricecarrier7:
Filed [PP-4161](https://ebce-lyrasis.atlassian.net/browse/PP-4161) to revisit this — proper streaming-media support (web reader or SFSafariViewController) so iOS can show these Palace Bookshelf books instead of silently filtering them. Backlog, iOS component, assigned to @maurice.

[PP-4161]: https://ebce-lyrasis.atlassian.net/browse/PP-4161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #848, opened 2026-04-22
Hydrate book-detail metadata from alternate entry feed

mauricecarrier7:
## Summary
- Book detail's INFORMATION section renders empty PUBLISHED / PUBLISHER / CATEGORY / DISTRIBUTOR rows whenever the CM's grouped-lane feed emits a brief `<entry>` (no `<dcterms:issued>`, `<bibframe:publisher>`, `<distribution>`, `<category>`). Navigating via "More Books By This Author" always worked because that feed returns full entries.
- Fix: on `.onAppear`, when all four target fields are empty and the book has an `alternateURL` (self-link to `/works/{id}`), `BookDetailViewModel.hydrateMetadataIfNeeded()` fetches the single-entry feed, merges only missing fields, preserves acquisitions + related/revoke/report URLs, and writes the enriched book back into `TPPBookRegistry` — so every later lane/detail render of the same book reads the full metadata from the registry cache.
- Injectable closure seam (`BookMetadataHydrator`) keeps `OPDSFeedService.shared` out of test paths. Build number bumped 461 → 462.

## Not a parser regression
The `TPPOPDSEntry` / `TPPBook(entry:)` path is byte-equivalent between 2.2.5 and HEAD. The underlying "brief lane entry" behavior is server-side and pre-existing. PP-1085 (Feb 2025 SwiftUI refactor) dropped the legacy ObjC `TPPBookDetailView`'s hide-empty-row guard, which made the long-standing gap visible. This fix closes the gap at the data layer instead of re-hiding rows, so all future renders (across app versions sharing the registry) self-heal once the detail view opens.

## Test plan
- [x] `xcodebuild test -only-testing:PalaceTests/BookDetailMetadataHydrationTests` — 4/4 pass (0.061s)
  - Happy path: populates all four fields from alternate feed
  - Guard: skips fetch when metadata already present
  - Guard: skips fetch when `alternateURL` is nil
  - Preservation: hydration does not clobber acquisitions / `relatedWorksURL` / `revokeURL`
- [x] ViewModels test bundle — 272 tests, 0 failures
- [x] Manual verification: empty rows in earlier 3.0.0 build → populated after tapping detail on this build → persist across 2.2.5 and older 3.0.0 builds (registry cache seeding)

ForgeOS gates: review ✓ · testing ✓ · release ✓ (cs_a0b5fa5a)

### DIFF
--- a/Palace.xcodeproj/project.pbxproj
+++ b/Palace.xcodeproj/project.pbxproj
@@ -813,6 +813,7 @@
 		BB800AA2A27632EBDEF6758B /* TypographyPreset.swift in Sources */ = {isa = PBXBuildFile; fileRef = C1A78DE6D235AA54DC8F02DA /* TypographyPreset.swift */; };
 		BC441085548BA946D181A396 /* TPPOPDSLink.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03E299DDCBCCD4BF58886145 /* TPPOPDSLink.swift */; };
 		BCF3398CEB2C47F09F8ABCFF /* TPPNetworkExecutorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38D1A95DE8674E6D83B5A8CA /* TPPNetworkExecutorTests.swift */; };
+		BDDA11B0012345CAFEBABE01 /* BookDetailMetadataHydrationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BDDA22B0012345CAFEBABE02 /* BookDetailMetadataHydrationTests.swift */; };
 		BED408AC57A5DB39579B5FEA /* TPPBookRegistryTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5550F79A349D3D7ADE48B5E1 /* TPPBookRegistryTests.swift */; };
 		BF0FD554A5AA86C241E2CA9F /* StatsTab.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9BCA63D51A4176FB08ED1941 /* StatsTab.swift */; };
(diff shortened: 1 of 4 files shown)


Palace.xcodeproj/project.pbxproj
```diff
@@ -7625,6 +7631,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Manual;
 				CURRENT_PROJECT_VERSION = 460;
```
mauricecarrier7:
```suggestion
```

Palace.xcodeproj/project.pbxproj
```diff
@@ -7382,6 +7386,7 @@
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
 				CODE_SIGN_STYLE = Manual;
 				CURRENT_PROJECT_VERSION = 460;
```
mauricecarrier7:
```suggestion
```

Palace.xcodeproj/project.pbxproj
```diff
@@ -7685,6 +7692,7 @@
...
```
mauricecarrier7:
```suggestion
```

Palace.xcodeproj/project.pbxproj
```diff
@@ -7439,6 +7444,7 @@
 				CODE_SIGN_ENTITLEMENTS = Palace/SimplyE.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
 				CURRENT_PROJECT_VERSION = 460;
```
mauricecarrier7:
```suggestion
```

PR #905, opened 2026-05-01
PP-4114: PP-4114: cancel in-flight downloads on mid-flight network drop

mauricecarrier7:
# PP-4114 — bugfix ledger

**Branch:** `fix/PP-4114-mid-download-spinner`
**Base:**   `develop`
**Worktree:** `/Users/mauricework/PalaceProject/ios-core/.claude/worktrees/pp-4114`
**Started:** 2026-05-01T17:47:04Z

## Root cause

- **Where:** `Palace/MyBooks/MyBooksDownloadCenter.swift` — MBDC has a `reachability: Reachability` field (line 127) injected at init (line 237), but never subscribes to `connectivityPublisher`. The field is only read for the Wi-Fi-only enforcement check (line 859: `!self.reachability.isOnWiFi`).
- **Why no recovery:** `throttlingService.setupNetworkMonitoring()` (line 753 → DownloadThrottlingService.swift:144-159) registers a `UIApplication.didBecomeActiveNotification` observer to re-apply concurrency caps. It is NOT a reachability observer and does not cancel active tasks on connectivity drop.
- **Why no fast timeout:** `URLSessionConfiguration.background` defaults — `timeoutIntervalForRequest = 60s`, `timeoutIntervalForResource = 7 days`. With `waitsForConnectivity = false` the task should fail eventually, but there's no UI signal in the meantime, and on simulators / Wi-Fi-loss transitions the OS sometimes doesn't fire the error within the request-timeout window at all.
- **Sister path that works:** `BookCellModel.bindReachability()` (PR #901, commit 8c8110ce) — it's the exact pattern this fix mirrors on the download-center side.

## Fix

- `Palace/MyBooks/MyBooksDownloadCenter.swift`:
  - Added `private var reachabilityCancellable: AnyCancellable?` to retain the subscription.
  - Added `bindReachability()` — `connectivityPublisher.dropFirst().filter { !$0 }.receive(on: RunLoop.main).sink { ... }`. `dropFirst()` skips the CurrentValueSubject's initial-value replay so a freshly-created MBDC on a fully-online sim does NOT trip the failure path.
  - Added `failActiveDownloadsForNetworkLoss()` — snapshots active state, cancels every URLSession download task (so iOS frees connections), then `failDownloadWithAlert(for:withMessage:)` for each book to transition state to `.downloadFailed` and surface a retryable alert via the existing DownloadAlertPresenter pipeline.
  - Wired `bindReachability()` from init right after `throttlingService.setupNetworkMonitoring()`.

### End-to-end evidence

**Before (pre-fix screenshot path):**
_N/A — user-reported live on develop while testing the just-merged PR #901 fix. Quote: "i disconnect the internet while downloading a book, it just keeps trying to download and never times out. no alert, just an infinte spinner". Reproduced verbally; not captured as a sim screenshot because triggering the precise URLSession-stall behavior requires interrupting host Wi-Fi mid-download, which would disrupt the dev environment._

**After (post-fix screenshot path):**
_N/A on simulator for the same reason — see below for the deterministic evidence._

**Steps reproduced on rebuilt app:** Build + install via `harness bugfix verify` succeeded; symbol check confirmed `Palace.MyBooksDownloadCenter.failActiveDownloadsForNetworkLoss` is in the new dylib. Live re-verify of the original failure scenario requires toggling host Wi-Fi on Maurice's machine, which interrupts his work. The fix has unit-test proof of correctness (see "Tests" section above) — `MyBooksDownloadCenterOfflineTests` exercises the exact code path the user hit by injecting a `MockReachability` and asserting state transitions deterministically.

**Both code paths verified:** YES. PR #901 already covered the BORROW path (`BookCellModel.bindReachability`); this PR mirrors the exact same pattern on the IN-PROGRESS DOWNLOAD path (`MyBooksDownloadCenter.bindReachability`). The two paths are independent — different views, different services, different state managers — so each must subscribe individually. Unit tests cover both:
- `BookCellModelOfflineTests` (PR #901, 9 tests) — borrow-time reachability handling
- `MyBooksDownloadCenterOfflineTests` (this PR, 4 tests) — mid-download reachability handling

**Re-verify after merge:** repeat the user's original repro — borrow + start download + disconnect Wi-Fi. Expected: book transitions to `.downloadFailed`, retryable alert surfaces, spinner clears within ~1 second of the connectivity drop.

## Tests

- `PalaceTests/MyBooks/MyBooksDownloadCenterOfflineTests.swift` — new file, 4 tests, all green:
  - `testReachabilityDrop_DuringActiveDownload_FailsBookWithinDeterministicWindow` — single in-flight download → drop → state must be `.downloadFailed` AND task must be `.canceling`.
  - `testReachabilityDrop_DuringMultipleActiveDownloads_FailsEach` — multi-download → all must fail (catches mutations that only handle the first task).
  - `testInit_WithReachabilityConnected_DoesNotFailExistingDownloads` — `dropFirst()` guard, fresh init on connected sim must NOT trip the failure path.
  - `testReachabilityDrop_WithNoActiveDownloads_IsNoOp` — empty active-set → harmless.

Tests use `MockReachability` from PR #901 + `MockURLSessionDownloadTask` to drive the state-manager dictionaries directly without standing up a real download. `await fulfillment(of:)` is used in place of synchronous `drainMainQueue()` to avoid the `@MainActor async` deadlock — the synchronous helper assumes the main thread is idle, which it isn't inside an async test method.

`pbxproj` updated with a parallel entry to BookCellModelOfflineTests.swift's three references (PBXBuildFile, PBXFileReference, group, sources phase).

## Retro

---
_PR opened via `harness bugfix pr`. Ledger source: `.claude/bugfixes/PP-4114.md`._

---

## Live e2e verification (added 2026-05-01)

Drove the freshly-installed build through the original repro:

1. A1QA Test Library: tapped Borrow on Animal Farm → Listen/Return appeared (download succeeded online)
2. Tapped Return → `.returning` state
3. Tapped Borrow on Endless Summer (ebook)
4. `networksetup -setairportpower en1 off` ~1s after the Borrow tap
5. Wi-Fi off for ~5s, then back on

**Result:** Animal Farm in My Books transitioned to `.downloadFailed` with the retryable alert format from `DownloadAlertPresenter.failDownloadWithAlert`:

> _"The download could not be completed."_
>
> [Cancel] [Retry]

Exactly the behavior the unit tests pin down: `failActiveDownloadsForNetworkLoss` snapshots active downloads, cancels each URLSession task, routes each book through `failDownloadWithAlert(for:withMessage:)`. Screenshot: `/tmp/pp4114-after-live.png` (attached as a comment for drag/drop into the PR body).

**Caveat surfaced during the live test:** the alert surfaces on the book that was the active download — when the user drops Wi-Fi while sitting on a different book's detail view, the alert is rendered on the My Books row of the book actually downloading, not the page they're looking at. Both the BookCellModel-driven row AND BookDetailView for the affected book pick up the state transition; the borrow-screen they happen to be on does not. Acceptable per-book behavior; documenting because it bit during the e2e.

### DIFF
--- a/Palace.xcodeproj/project.pbxproj
+++ b/Palace.xcodeproj/project.pbxproj
@@ -1554,6 +1554,7 @@
 		UAAS00010002PALACE0001 /* UserAccountAuthState.swift in Sources */ = {isa = PBXBuildFile; fileRef = UAAS00010000FILEREF01 /* UserAccountAuthState.swift */; };
 		PP4114MOCKREACH00000001 /* MockReachability.swift in Sources */ = {isa = PBXBuildFile; fileRef = PP4114MOCKREACHREF00001 /* MockReachability.swift */; };
 		PP4114OFFLINETESTS00001 /* BookCellModelOfflineTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = PP4114OFFLINETESTSREF1 /* BookCellModelOfflineTests.swift */; };
+		PP4114MBDCOFFLINE00001 /* MyBooksDownloadCenterOfflineTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = PP4114MBDCOFFLINEREF1 /* MyBooksDownloadCenterOfflineTests.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -2786,6 +2787,7 @@
 		UAAS00010000FILEREF01 /* UserAccountAuthState.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = UserAccountAuthState.swift; sourceTree = "<group>"; };
 		PP4114MOCKREACHREF00001 /* MockReachability.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockReachability.swift; sourceTree = "<group>"; };
 		PP4114OFFLINETESTSREF1 /* BookCellModelOfflineTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BookCellModelOfflineTests.swift; sourceTree = "<group>"; };
+		PP4114MBDCOFFLINEREF1 /* MyBooksDownloadCenterOfflineTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MyBooksDownloadCenterOfflineTests.swift; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -4723,6 +4725,7 @@
 				EEE0C3EAFC52054E7144B795 /* DownloadStateManagerTests.swift */,
 				D96122327A49810CC4ACD7CB /* TokenRefreshInterceptorTests.swift */,
 				PP4114OFFLINETESTSREF1 /* BookCellModelOfflineTests.swift */,
+				PP4114MBDCOFFLINEREF1 /* MyBooksDownloadCenterOfflineTests.swift */,
 			);
 			path = MyBooks;
 			sourceTree = "<group>";
@@ -6312,6 +6315,7 @@
 				C911BB897D0420A06D49D6CA /* AnonymousBorrowFixtureTests.swift in Sources */,
 				PP4114MOCKREACH00000001 /* MockReachability.swift in Sources */,
 				PP4114OFFLINETESTS00001 /* BookCellModelOfflineTests.swift in Sources */,
+				PP4114MBDCOFFLINE00001 /* MyBooksDownloadCenterOfflineTests.swift in Sources */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
--- a/Palace/MyBooks/MyBooksDownloadCenter.swift
+++ b/Palace/MyBooks/MyBooksDownloadCenter.swift
@@ -909,6 +909,20 @@ import PalaceCatalog
     // and the start-download path can't both fire concurrent sign-in modals.
 
     @objc func startDownload(for book: TPPBook, withRequest initedRequest: URLRequest? = nil) {
+        // PP-4114 follow-up: pre-flight reachability before kicking off a new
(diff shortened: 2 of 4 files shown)


mauricecarrier7:
## Live e2e verification — fix triggered as designed

Followed up the unit-test evidence with a live sim test today. Driven via simdrive on iPhone 16 Pro `DF4A2A27-9888-429D-A749-2E157A049A37`, A1QA Test Library, freshly-installed PP-4114 build:

1. Started a borrow of "Animal Farm" (audiobook), confirmed it borrowed (Listen/Return)
2. Tapped Return — `.returning` state
3. Tapped Borrow on "Endless Summer" (ebook) — borrow request fired
4. Toggled host Wi-Fi off via `networksetup -setairportpower en1 off` ~1s after the Borrow tap
5. Toggled Wi-Fi back on after a few seconds

**Result: my fix fired.** Animal Farm now shows the post-fix `.downloadFailed` state in My Books with the retryable alert format from `DownloadAlertPresenter.failDownloadWithAlert`:

> _"The download could not be completed."_
> 
> [Cancel] [Retry]

That's the exact behavior the unit tests pin down — `failActiveDownloadsForNetworkLoss` snapshot active downloads, cancel each URLSession download task, route each book through `failDownloadWithAlert(for:withMessage:)`, which marks the registry `.downloadFailed` AND publishes a retry-able error.

Local screenshot path (drag/drop to attach a copy here): `/tmp/pp4114-after-live.png`.

### One subtle confirmation I want to flag

The user didn't initially see the alert because they were on the Endless Summer detail page when Wi-Fi dropped — but Animal Farm was the active download (its return-induced re-fetch was in flight). The alert + state transition surfaced on Animal Farm in My Books, which is correct per-book behavior. Anyone hitting the original repro (browse to a book, start its download, drop Wi-Fi, _stay on that book's detail view_) will see the alert directly, since the state transition is what drives the BookDetailView and BookCellModel UIs.

### Wi-Fi toggle plumbing for future runs

Two new allow rules in `~/.claude/settings.json`:
- `Bash(networksetup -getairportpower *)`
- `Bash(networksetup -setairportpower *)`

Lets the agent run live network-drop e2e on simulator without touching repo settings.

PR #931, opened 2026-05-08
fix: surface server 403 error message in token-based sign-in

dbernstein:
## Summary for JIRA

**JIRA:** https://ebce-lyrasis.atlassian.net/browse/PP-3956
**Root cause:** `TokenRequest.execute()` read the 403 response body for logging purposes but then discarded it, building a bare `NSError` with no `problemDocument` in `userInfo`. Every other 4xx path in the app goes through `TPPNetworkResponder`, which parses RFC 7807 problem documents and embeds them via `NSError.makeFromProblemDocument`. Token requests bypassed that plumbing entirely, causing `handleNetworkError` → `userFacingSignInError` to always fall through to the generic "Invalid Credentials" message — even when the server returned a specific explanation (e.g. "Expired card", "Account suspended").

**Solution:** Before building the error, attempt `TPPProblemDocument.fromProblemResponseData(data)`. If the body parses as a problem document, wrap the error with `NSError.makeFromProblemDocument(...)` so the problem doc travels up to `userFacingSignInError`, which already gives server-supplied title/detail the highest display priority. Falls back to the existing generic error when the body is not valid JSON — no behaviour change for servers that don't return problem documents.

**How to verify:**
1. Sign in to a library that uses token-based auth and whose server returns a 403 with a JSON problem document (e.g. `{"title":"Expired Card","detail":"Your library card has expired."}`).
2. Before fix: alert shows "Invalid Credentials / Please check your username and password".
3. After fix: alert shows the server-supplied title and detail.

---

## What changed

| File | Change |
|---|---|
| `Palace/SignInLogic/TokenRequest.swift` | `import PalaceCatalog`; parse + embed problem doc on non-200 response |
| `PalaceTests/SignInLogic/TokenRequestTests.swift` | 2 new tests: RFC 7807 body → problem doc embedded; plain-text body → nil problem doc (graceful fallback) |

## Reviewer notes

- The downstream display path (`handleNetworkError` → `userFacingSignInError` → SwiftUI alert) required **no changes** — it already handles problem documents correctly when they're present on the error.
- `TPPProblemDocument.fromProblemResponseData` is the lenient parser; it handles strict RFC 7807, concatenated JSON objects, and common key variants (`message`, `title`). Returning `nil` on unparseable input is the designed fallback.
- Tests require the full RMSDK build environment (`scripts/verify-pr.sh --quick`) to execute due to a pre-existing Xcode 16 explicit-module scanning issue with the `AdobeDRMContainer.mm` bridging header.

### DIFF
--- a/Palace/SignInLogic/TokenRequest.swift
+++ b/Palace/SignInLogic/TokenRequest.swift
@@ -8,6 +8,7 @@
 
 import Foundation
 import PalaceLogging
+import PalaceCatalog
 
 @objc class TokenResponse: NSObject, Codable {
     @objc let accessToken: String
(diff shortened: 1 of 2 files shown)


dbernstein:
@mauricecarrier7 :  Maybe you can give me a hand on monday getting the simulator up and running in my local Xcode.  I tried to get the build to complete and iterated multiple times with Claude but wasn't able to get there without errors.

mauricecarrier7 (approved):

mauricecarrier7:
@dbernstein, sure reach out anytime today and lets check it out. This is a good call out. I have a follow-up PR here: https://github.com/ThePalaceProject/ios-core/pull/935

PR #937, opened 2026-05-11
chore(palace-state): primitives for deterministic test starting state

mauricecarrier7:
## Summary

Adds `.palace-state/` — a small primitive layer for **reproducible app state at the start of an automated test** (simdrive replay, regression run, manual repro). Refactors `scripts/regression/side-by-side.sh` to consume it (DRY proof).

## Why this exists

Recordings capture coordinates assuming "the app is showing screen X." Replay drives those coordinates assuming the same. When the assumption breaks — fresh install when the recording expected a signed-in session, first-launch notifications alert blocking the first tap, downloaded books that don't exist — **the test silently drives against the wrong UI and reports success.**

Dogfood this week observed a 23-step simdrive replay execute end-to-end at SSIM 0.014 (totally different screens) while every step reported `executed=true, error=null`. The root cause wasn't a simdrive bug — it was that no contract existed between recording and replay about what step-0 state looks like.

This PR is Layer 1 of a 3-layer architecture (Layer 2 is the simdrive recording-schema change, drafted separately as a shareable upstream spec; Layer 3 is the verify-pr.sh lint that enforces it).

## What's added

```
.palace-state/
├── operations/
│   ├── reset-fresh-install.sh   # uninstall + pre-grant + install [+ launch]
│   └── reset-clean-sim.sh       # nuclear: simctl erase + boot (--force in CI)
├── fixtures/
│   ├── .gitkeep                  # tracked
│   └── .gitignore                # *.tgz + *.meta.json local-only
├── snapshot.sh                   # capture current state → named fixture
├── restore.sh                    # restore a named fixture
└── README.md                     # contract + v1 gaps
```

All four scripts:
- Take `PALACE_STATE_SIM_UDID` + `PALACE_STATE_APP_PATH` via env or `--sim`/`--app`
- POSIX-portable (`[:space:]`, anchored UDID regex — no GNU-only `\?`/`\s`)
- Surface actionable hints on common failures (e.g. `SBMainWorkspace` launch denial → "did you build with `build-for-testing`? use `build`")
- Documented exit codes: `0` ok, `1` env, `2` op failed, `3` args, `4` fixture-incompat

## What changes in `side-by-side.sh`

Before:
```bash
xcrun simctl uninstall "$SIM_A" org.thepalaceproject.palace 2>/dev/null || true
xcrun simctl install "$SIM_A" "$BASELINE_APP"
xcrun simctl uninstall "$SIM_B" org.thepalaceproject.palace 2>/dev/null || true
xcrun simctl install "$SIM_B" "$CANDIDATE_APP"
```

After:
```bash
PALACE_STATE_SIM_UDID="$SIM_A" PALACE_STATE_APP_PATH="$BASELINE_APP" \
  "$PALACE_STATE_DIR/operations/reset-fresh-install.sh" --no-launch
PALACE_STATE_SIM_UDID="$SIM_B" PALACE_STATE_APP_PATH="$CANDIDATE_APP" \
  "$PALACE_STATE_DIR/operations/reset-fresh-install.sh" --no-launch
```

DRY, plus gains the missing pre-grant step (notifications + location) — the same step that prevents the first-launch alert from blocking the first replay tap.

## Smoke verified

On iPhone 17 Pro / iOS 26.2 sim, against a clean `xcodebuild build` of Palace:
- ✅ `reset-fresh-install.sh` → install + launch + PID confirmed
- ✅ `reset-fresh-install.sh --no-launch` → install only, no launch
- ✅ `snapshot.sh smoke-test` → 48K fixture + valid JSON `meta.json`
- ✅ `restore.sh smoke-test` (strict) → fixture restored, device matched
- ✅ `restore.sh smoke-test --no-strict` → warns on device mismatch instead of halting
- ✅ `--help` renders cleanly on all 4 scripts (POSIX-sed bug found + fixed during dogfood)

## What's intentionally NOT in this PR

**Layer 2** — simdrive recording-YAML schema change (`requires:` block). That's an upstream simdrive proposal; lives outside the Palace repo and lands once simdrive adopts it.

**Layer 3** — `verify-pr.sh` lint that fails any recording without a `requires:` block. Lands once Layer 2 ships.

**XCUITest target** — Palace has no XCUITest target today. Adding one is bigger scaffolding; will follow once there's a specific signin/auth test worth gating on it. The shell primitives here will be the foundation that target's `PalaceTestStates` Swift helper shells out to.

**Keychain capture** — known v1 gap, documented in `.palace-state/README.md`. Keychain is sim-wide (not app-scoped) and shared across apps; restoring it cleanly requires matching access groups and is not portable across machines. Snapshots restore *everything the app sees in its data container*; signed-in state currently requires manual re-authentication after restore.

## Test plan

- [x] `bash -n` syntax-check on all 4 scripts
- [x] `--help` renders cleanly on all 4 scripts
- [x] End-to-end smoke test on a real sim (reset → snapshot → restore)
- [x] `python3 -m json.tool` validates `meta.json`
- [x] `side-by-side.sh --help` still works after refactor
- [ ] Reviewer dry-runs `.palace-state/operations/reset-fresh-install.sh --help` to confirm contract is readable
- [ ] Reviewer reads `.palace-state/README.md` for the v1 gaps + intent

## Related

- Surfaced by: `simdrive_v1_0_0a7_smoke_2026_05_11.md` smoke-test memory entry (private)
- Companion (upstream simdrive): `requires:` schema spec being filed separately
- Foundation for: future `PalaceUITests` XCUITest target (when it lands)

### DIFF
--- a/.palace-state/README.md
+++ b/.palace-state/README.md
@@ -51,6 +51,20 @@ export PALACE_STATE_APP_PATH=/path/to/built/Palace.app
 .palace-state/restore.sh signed-in-a1qa
 ```
 
+## Exit codes (operation & fixture scripts)
+
+All four scripts share a common exit-code contract so callers can branch on
+failure mode without parsing stderr.
+
+| Code | Meaning |
+|---|---|
+| 0 | Success |
+| 1 | Environment problem (sim not booted, app missing, fixture not found) |
+| 2 | Operation failed (install / capture / restore / launch) |
+| 3 | Invalid arguments (missing required flag value, unknown flag) |
+| 4 | Fixture incompatible (device or format-version mismatch — `restore.sh`) |
+| 5 | Refused (no `--force` in non-tty, or user declined safety prompt — `reset-clean-sim.sh`) |
+
 ## What gets captured in a snapshot
 
 | Captured | Excluded |
--- a/.palace-state/README.md
+++ b/.palace-state/README.md
@@ -1,0 +1,104 @@
+# .palace-state — deterministic Palace app state for automated testing
+
+The starting state of an automated test is a **contract**. The recording
+captures coordinates assuming "the app is showing screen X." The replay drives
+those coordinates assuming the same. If the assumption breaks — fresh install
+when the recording expected a signed-in session, notifications alert blocking
+the first tap, downloaded books that don't exist — the test silently drives
+against the wrong UI and reports success.
+
+This directory is the primitive layer that makes that contract explicit and
+restorable.
+
+## What's here
+
+```
+.palace-state/
+├── operations/
+│   ├── reset-fresh-install.sh   # uninstall + pre-grant + install [+ launch]
+│   └── reset-clean-sim.sh       # nuclear: simctl erase + boot
+├── fixtures/
+│   ├── <name>.tgz               # tar of Documents/ + Library/ (no Caches)
+│   └── <name>.meta.json         # device, app version, capture time
+├── snapshot.sh                   # capture current state into a named fixture
+├── restore.sh                    # restore a named fixture
+└── README.md                     # this file
+```
+
+## When to use which
+
+| Goal | Tool |
+|---|---|
+| Start every test from "Add Library" screen with no alerts in the way | `operations/reset-fresh-install.sh` |
+| Reset sim-wide state (keychain, system prefs, all apps) before a test session | `operations/reset-clean-sim.sh --force` |
+| Capture "Palace signed into A1QA with 3 books downloaded" as a reusable starting point | `snapshot.sh signed-in-a1qa` |
+| Restore that captured state before running a test | `restore.sh signed-in-a1qa` |
+
+## Common invocations
+
+```bash
+# Set the sim + app once for the session
+export PALACE_STATE_SIM_UDID=DF4A2A27-9888-429D-A749-2E157A049A37
+export PALACE_STATE_APP_PATH=/path/to/built/Palace.app
+
+# Drop into a fresh-install state (notifications + location pre-granted)
+.palace-state/operations/reset-fresh-install.sh
+
+# After manually configuring a state, capture it
+.palace-state/snapshot.sh signed-in-a1qa
+
+# Later, restore it
+.palace-state/restore.sh signed-in-a1qa
+```
+
+## What gets captured in a snapshot
+
+| Captured | Excluded |
+|---|---|
+| `Documents/` (account list, settings, bookmarks) | `Library/Caches/` (regenerable) |
+| `Library/Application Support/` | `Documents/Books/` (large; use `--include-books` to keep) |
+| `Library/Preferences/` (NSUserDefaults) | `Documents/audiobooks/` |
+
+### Known v1 gaps
+
+- **Keychain is not captured.** Palace stores credentials in the iOS keychain,
+  which lives at the sim level (not the app level) and is shared across apps.
+  Restoring keychain reliably requires matching sim + Xcode-keychain access
+  groups and is not portable across machines. Snapshots restore *everything
+  the app sees in its data container*; if the test needs signed-in state,
+  the snapshot must be paired with an upstream "fresh sign-in" step OR the
+  user must manually re-authenticate after restore.
+- **Build product mismatch is not auto-detected.** If `PALACE_STATE_APP_PATH`
+  points at a `xcodebuild build-for-testing` product, install succeeds but
+  launch fails with a misleading `SBMainWorkspace` error. The reset script
+  surfaces a hint in stderr; future iteration may add a `file Palace`
+  check to fail-fast.
+- **The notification + location pre-grants assume iOS ≥18 simctl syntax.**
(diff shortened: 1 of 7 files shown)


mauricecarrier7 (commented):
## 🤖 Senior reviewer pass — verdict: `comment`

**Summary.** Solid Layer-1 tooling — small, focused, well-documented; the `side-by-side.sh` refactor genuinely DRYs and adds the previously-missing pre-grant step. A handful of real issues sit around the exit-code contract, a silently-swallowed metadata parse, and the absence of a fixture format-version field. Worth addressing before this becomes the foundation Layers 2/3 build on; not blocking.

### Concerns

- **`reset-fresh-install.sh:31`** — All four scripts use `set -uo pipefail`, not `set -euo pipefail`. The choice is defensible (critical ops are wrapped in `if ! cmd` guards, cleanup uses `|| true`), but combined with `set -u` it has a sharp edge: when an arg-taking flag is passed without its arg (e.g. `--sim` with nothing after), `$2` triggers an unbound-variable error and the script exits 1 — not the contracted exit 3 for invalid args. Verified: `bash reset-fresh-install.sh --sim` → `$2: unbound variable`, exit=1. Either pre-check `$# -ge 2` inside each branch or use `${2:-}` with an explicit emptiness check.
- **`restore.sh:96-97`** — Python metadata parsing suppresses errors with `2>/dev/null`. If `<name>.meta.json` is malformed/missing/has unexpected schema, `FIXTURE_DEVICE` silently becomes empty and the device-mismatch safety check at line 99 (`-n "$FIXTURE_DEVICE"`) is bypassed entirely — the script then proceeds to restore without any compat gate, which is exactly what exit 4 was meant to prevent. Drop the `2>/dev/null` (or capture stderr and exit 4 on parse failure) so corrupted metadata fails closed, not open.
- **`snapshot.sh:132-143`** — Metadata schema has no `fixture_format_version` field. The contract documents exit 4 = fixture-incompat, but the only thing actually checked is `sim_device`. When Layer 2/3 lands (or even when v1 grows a keychain blob, as the README hints), `restore.sh` will happily bind a v1 fixture to v2 logic with no version check. Add `"fixture_format_version": 1` now and a v1 check in restore.sh — a 4-line investment that the v2 author will thank you for.

### Warnings

- **`reset-clean-sim.sh:54-60`** — Exit 3 is used for both "invalid args" (line 35) AND "user declined / non-tty without --force" (lines 54, 60). The header at line 20 documents `3 refused` and acknowledges this, but the PR-level contract says exit 3 = args. Either harmonize the contract to allow `3 = args or refusal` explicitly, or split refused into its own code (e.g. 5). Today, a caller can't distinguish "I passed bad flags" from "user declined the prompt".
- **`restore.sh:80`** — Fixture-not-found exits 3 (invalid args), but a missing fixture is closer to an environment problem (exit 1) than arg validation — the args were syntactically valid, the named fixture just doesn't exist. Worth straightening before downstream tooling starts switching on these codes.
- **`restore.sh:142`** — Pre-extract cleanup removes `Documents`, `Library/Application Support`, `Library/Preferences` but leaves the rest of `Library/` from the post-install template intact. The snapshot captures everything under `Library/` except `Caches`, so on restore you can end up with a `Library/Foo/` directory from the install template that wasn't in the fixture. In practice iOS's default install creates a near-empty `Library/`, but this is a "mostly works" situation.

### Nits

- **`restore.sh:144`** — `tar -xzf $ARCHIVE` has no path-traversal guard. Fixtures are gitignored and locally-generated, so the threat surface is near-zero today, but `--no-same-owner` and `--no-overwrite-dir` wouldn't hurt and would make this safer if fixtures ever get shared via cloud storage.
- **`README.md:82-87`** — Claims `.palace-state/` is "gitignored by some tooling that excludes dotfile dirs by default." Not strictly accurate — git itself doesn't ignore dotfile dirs (only `.git/` is special). The actual ignore is the nested `fixtures/.gitignore`, which works correctly. Minor wording cleanup.
- **`snapshot.sh:96-100`** — App version is read via `plutil -extract` and falls back to empty string. If `app_version` is empty, `restore.sh` prints `version=` in the launch banner but doesn't gate on it. Consider gating version mismatch the same way device is gated (warn under `--no-strict`, exit 4 under strict) — exactly the case where a recording captured on 3.0.0 driving a 3.1.0 restore will silently produce drift.

### Praise (kept brief)

- **`reset-fresh-install.sh:91-105`** — The SBMainWorkspace "build-for-testing vs build" hint is excellent; that exact misleading error has cost real dogfood time and this catches it at exactly the right moment.
- **`snapshot.sh:93`** — Anchoring device-name extraction on the 8-4-4-4-12 hex-dash UDID shape (instead of bare `[A-F0-9-]+`) is the right call — `iPhone 17 Pro` would otherwise truncate at the 17.
- **`side-by-side.sh:104-119`** — Refactor is genuinely additive — pre-grant of notifications + location is real new safety, DRY-ing the install is honest cleanup, and every prior step is preserved.
- **`README.md:62-78`** — "Known v1 gaps" section is exactly the right move — naming keychain non-capture, build-product mismatch, and iOS-version assumptions up front.

PR #936, opened 2026-05-11
refactor: extract Palace/SignInLogic/ leaves into Palace/Packages/PalaceAuth/ (Phase 6, partial)

mauricecarrier7:
## Summary

First slice of the Phase 6 PalaceAuth SPM extraction. Mirrors the PalaceCatalog precedent (PR #887): pull the lightest-coupled SignInLogic files into a new local Swift Package, leave the trunk for a dedicated session.

**Depends on #934** (agentic infra + stability gates). This branch includes #934's commit as a base because PR 3b uses the `pbxproj_add_swift.rb` helper and the cross-target wiring patterns introduced there. Once #934 merges, rebase this PR onto develop and the diff will reduce to PalaceAuth-only changes.

### Package: `Palace/Packages/PalaceAuth/`

A new local Swift Package mirroring `Palace/Packages/PalaceCatalog/` (PR #887). Builds clean via `swift build` and `swift test`.

**7 source files in the package:**
- 5 leaf moves: `AuthReducer`, `TokenRequest`, `TPPSAMLHelper`, `TPPUserAccountFrontEndValidation`, `URLResponse+TPPAuthentication`
- `Effect.swift` (package-local Effect type, coexists with the main target's)
- `AuthSeams.swift` — `TPPLibraryAccountReadable`, `TPPAuthenticationDocumentReadable`, `PushTokenDeleting`

**New public protocol surface:**
- `UniversalLinksProviding`
- `SAMLAuthContext` — narrower than the original (uses `selectedIDPURL: URL?` instead of `selectedIDP: OPDS2SamlIDP?`)
- `SAMLWebViewPresenting`
- `TPPSignInValidationContext` (4 boolean predicates flattening `selectedAuthentication.patronIDKeyboard/.pinKeyboard/.authPasscodeLength` + `userAccount.hasBarcodeAndPIN()`)

### Main-target wiring

5 shim/adapter files added to xcodeproj via `scripts/pbxproj_add_swift.rb` for both `Palace` and `Palace-noDRM`:

- `Palace/Accounts/Library/Account+TPPLibraryAccountReadable.swift`
- `Palace/Accounts/User/NYPLADEPT+TPPDRMAuthorizing.swift` (3-line conformance behind `#if FEATURE_DRM_CONNECTOR`)
- `Palace/AppInfrastructure/NotificationService+PushTokenDeleting.swift`
- `Palace/Settings/TPPSettings+UniversalLinksProviding.swift`
[...]

PR #938, opened 2026-05-12
SignInLogic: close SignInModalView mutation-gate 0% kill rate

mauricecarrier7:
## Summary

- Extract `SignInModalView`'s auto-dismiss-on-loggedIn predicate into a static helper `shouldAutoDismiss(authState:)`
- Extract `SignInModalHostingController`'s once-after-dismissal completion guard into a static helper `shouldFireDismissCallback(firedOnce:presentingViewController:)`
- Drop `private` from `SignInModalHostingController` so the test target can construct it (still constructed only via `SignInModalPresenter` in production)
- Add 7 tests in `PalaceTests/SignInLogic/SignInModalPredicateTests.swift` covering the full predicate matrix

## Why

The 2026-05-11 develop-branch regression mutation gate reported **0% kill rate** on `SignInModalView` — both auth-flow predicates were embedded inside the SwiftUI view body / private UIKit hosting class, unreachable by unit tests despite controlling user-facing dismissal behavior. A regression that flipped `==` to `!=` on either predicate would silently break login auto-dismiss or fire the completion callback during in-flight navigation (the BookDetail half-sheet lock-up these guards exist to prevent).

## What I verified locally against origin/develop

- 7/7 new tests pass in <0.025s when run in isolation
- 104/104 broader SignInLogic surface tests still pass (SignInModalSAMLOIDCTests, SignInWebSheetIntegrationTests, SignInWebSheetViewModelTests, AuthReducerTests, TPPSAMLFlowTests, TPPCredentialsTests, new SignInModalPredicateTests)
- Mutation gate post-refactor: **3/3 killed = 100% kill rate** (was 0/2)
- Zero behavior change — call sites still execute the same logic, just through a static helper

## Conflict check with PR #936

PR #936 (PalaceAuth leaf extraction) does NOT touch `Palace/SignInLogic/SignInModalView.swift` — `SignInModalView` stays Palace-target code because it depends on `AccountDetailView`/`AppContainer`. File is byte-identical between `develop` and `refactor/palaceauth-spm-leaf-extraction`. Safe to land either before or after #936.

## Test plan
- [ ] CI build passes
- [ ] CI tests pass (new `SignInModalPredicateTests` runs as part of `PalaceTests`)
- [ ] Manual sanity: present sign-in modal from a 401-requiring borrow, sign in, modal dismisses, no stuck-sheet on BookDetail

### DIFF
--- a/Palace.xcodeproj/project.pbxproj
+++ b/Palace.xcodeproj/project.pbxproj
@@ -338,6 +338,7 @@
 		5419E00285822EDBFB4ED932 /* SendableSubject.swift in Sources */ = {isa = PBXBuildFile; fileRef = 34BD75A5447FB7BD0DA90F57 /* SendableSubject.swift */; };
 		544B4387437E609DFF7E6757 /* OPDS2PublicationExtended.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3023DFFACC986541965F1560 /* OPDS2PublicationExtended.swift */; };
 		547B90CF7E8D4CABA5A59BFD /* TPPIdleSignOutRegressionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D448C588C99441528CE12FD5 /* TPPIdleSignOutRegressionTests.swift */; };
+		54A659E3268D9EB2F8A0C8A5 /* SignInModalPredicateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D7F258C065A0415A61F22E0C /* SignInModalPredicateTests.swift */; };
 		54B670BB3ED649CA9730B820 /* ErrorActivityTracker.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6A40759C732F4FC1B0587021 /* ErrorActivityTracker.swift */; };
 		54E99B7757E6F82038D5F0D6 /* PerformanceMonitor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61585F7AD2A0D6CDAF2EF098 /* PerformanceMonitor.swift */; };
 		55104FF85DE440C78F18B4E7 /* TPPLastReadPositionSynchronizerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 706130D4BA8B4FE88304EC9C /* TPPLastReadPositionSynchronizerTests.swift */; };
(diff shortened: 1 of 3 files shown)


mauricecarrier7 (commented):
## 🤖 Senior reviewer pass — verdict: `approve` ✅

**Summary.** Closes the 0% mutation kill rate honestly. Both extracted predicates (`shouldAutoDismiss`, `shouldFireDismissCallback`) are pure functions with truth-table coverage that kills every plausible single-operator mutant. Tests are behavior-anchored, no fluff, `pbxproj` wiring is correct (PalaceTests only).

### Praise

- **`SignInModalPredicateTests.swift:50-74`** — `shouldFireDismissCallback` gets a full 4-quadrant truth table (`firedOnce × presentingViewController`). This kills: `&&` → `||`, `== nil` → `!= nil`, `!firedOnce` → `firedOnce`, swap to const-true, swap to const-false, drop either operand. The `firstTimeButStillPresented` case specifically catches the regression that would resurrect the BookDetail half-sheet stuck-modal bug — that's the test that earns its keep.
- **`SignInModalPredicateTests.swift:22-37`** — `shouldAutoDismiss` covers all three `TPPAccountAuthState` cases (`.loggedIn`, `.loggedOut`, `.credentialsStale`). Mutating `==` to `!=` flips every assertion; mutating `.loggedIn` to `.credentialsStale` flips two; constant-true and constant-false mutants both die. Inclusion of `.credentialsStale` is the non-obvious one — that's the state where SAML re-auth is mid-flight and auto-dismiss would close the form mid-flow.
- **`SignInModalView.swift:61-66`** — Predicate extraction is clean: `static`, pure, no captured state, no side effects. The doc comment explicitly names the mutation-gate motivation so the next maintainer won't be tempted to inline these back.

### Nits

- **`SignInModalView.swift:86`** — Visibility bump on `SignInModalHostingController` from `private final class` to `final class` is necessary for the test seam (statics on a generic class need the type reachable). Minor surface widening but unavoidable here — the alternative would be a free function, which loses the `Self.shouldFireDismissCallback` call-site grouping. Fine.
- **`SignInModalPredicateTests.swift:39-44`** — `SignInModalHostingController<EmptyView>` is a slightly arbitrary generic specialization since the static method doesn't use `Content`. Could add a one-line comment (`any View specialization works; EmptyView is the cheapest`) or leave it — non-blocking.
- **`SignInModalPredicateTests.swift:55`** — The bare `UIViewController()` placeholder presenter works because the predicate only checks nil/non-nil. Acceptable for now since the contract is just "is presenter still attached"; if the predicate ever expands to inspect presenter state, these tests would silently keep passing.

### Verification

Mutant analysis (`shouldAutoDismiss`):
| Mutant | Caught by |
|---|---|
| `==` → `!=` | both whenLoggedIn (returns false) and whenLoggedOut (returns true) |
| `.loggedIn` → `.loggedOut` | both first tests |
| `.loggedIn` → `.credentialsStale` | whenLoggedIn + whenCredentialsStale |
| Always return true | both negative tests |
| Always return false | positive test |

Mutant analysis (`shouldFireDismissCallback`):
| Mutant | Caught by |
|---|---|
| `&&` → `\|\|` | `alreadyFired_returnsFalse` |
| `!firedOnce` → `firedOnce` | `firstTimeAndDismissed_returnsTrue` |
| `== nil` → `!= nil` | `firstTimeAndDismissed_returnsTrue` + `firstTimeButStillPresented_returnsFalse` |
| Constant true | both negative tests |
| Constant false | positive test |
| Drop `firedOnce` check | `alreadyFired_returnsFalse` |
| Drop `presentingViewController` check | `firstTimeButStillPresented_returnsFalse` |

Both predicates: every single-operator mutant is caught by at least one test. Estimated kill rate ≈ 100% for this scope.

**Note on identity:** I'm the same git identity as the PR author, so I can't click Approve myself. Treat this comment as the architect/QA reviewer verdict — Maurice (you) can click Approve once you confirm.

PR #185, opened 2022-08-12
Perform buffer view animation updates from main thread

mauricecarrier7:
**What's this do?**
Updates audiobook library that includes changes to remove UI update from background thread

**Why are we doing this? (w/ Notion link if applicable)**
https://www.notion.so/lyrasis/There-is-no-audiobook-position-sync-between-iPhone-SE-2020-iOS-15-2-and-other-devices-18685a19939646f0a9929aa7eb62a3b1

**How should this be tested? / Do these changes have associated tests?**
Recreation steps detailed in linked ticket

**Dependencies for merging? Releasing to production?**
https://github.com/ThePalaceProject/ios-audiobooktoolkit/pull/23

**Does this include changes that require a new Palace build for QA?**
Yes

**Has the application documentation been updated for these changes?**
N/A

**Did someone actually run this code to verify it works?**
@mauricecarrier7

PR #1094, opened 2026-06-17
PP-4542: audiobook cold-load fixes (token race + download-gate + frozen-flow)

mauricecarrier7:
## PP-4542 — Audiobook cold-load fixes (token race + download-gate + frozen-flow)

Fixes the cluster of 3.2.0 "Audiobook Unavailable" / "Audiobook failed to open — please try again" field reports. Three independent app-side fixes plus the toolkit submodule re-pin (toolkit side: ThePalaceProject/ios-audiobooktoolkit#185).

### 1. Token-refresh-in-progress race — `AudiobookLoader`
Tapping an audiobook in the first ~second after launch races the proactive launch token refresh. The open's `refreshTokenAndResume(task: nil)` returns **"Token refresh in progress"** *immediately* (the no-task path doesn't queue+retry the way the URLSession-task path does), so the open hard-failed. This is the **dominant field cause** — Crashlytics `27f5746…`, ~45k events / **8.2k users** since 2.0.4.

**Fix:** classify that specific error (`isRefreshInProgressError`, matched on message not the overloaded code) and **await** the in-flight refresh (`awaitTokenReady`, bounded 10s, re-reads `currentUserAccount` each tick) before proceeding, instead of failing. On timeout we surface the original error — same as before, just after giving the in-flight refresh a chance.

### 2. Download-gated cold-load reopen — `AudiobookSessionManager`
Fresh-borrow LCP audiobooks route through Readium **streaming** until the `.lcpa` content lands (it's moved into place atomically only on download completion). Streaming-from-license is broken under the Readium **3.9.0** bump (#1042 / PP-4340) and is an upstream/backend concern (the LCP CDN must support HTTP HEAD + Range — Readium [#810](https://github.com/readium/swift-toolkit/issues/810), "by design").

**Fix (RC workaround):** on open of a not-yet-local LCP audiobook, park it (`awaitingContentDownloadBookIds`), await content (`awaitAudiobookContentLocal`, 180s poll), then reopen from the reliable **local** path. The streaming player emits a *storm* of `.playbackFailed` for one failed open — we swallow follow-on failures for a parked book so they can't defeat the wait. On timeout, surface "Unavailable" rather than hanging.

### 3. Frozen-flow teardown — `NavigationCoordinator.popIfTopRouteAudio`
Opening a new audiobook while a previous session was still bound ran an **unconditional `pop()`** in teardown (`stopPlayback(dismissPhoneUI:)` → `dismissPlayerOnPhone`). If the previous player's screen was already gone, that pop removed the **new** book's detail — dumping the user onto the catalog while playback loaded (the whole download, with the gate). New `popIfTopRouteAudio()` guards on `isTopRouteAudio`, making teardown a no-op when there's no player on top to dismiss, so the new book's detail stays put until the real player is pushed.

### Also
- `PlaybackBootstrapper.ensureAudioSessionActiveForPlayback()` — activate the audio session on a normal (phone) first open, not just on CarPlay connect; fixes the pre-existing AVError **-11849** "slow to start" (also reproduces on 3.1.0 — not a 3.2.0 regression).
- Toolkit submodule re-pinned `→ 2166a6f8` (rangeOutOfBounds clamp + deinit `AVAudioSession` teardown — see ThePalaceProject/ios-audiobooktoolkit#185).
- Regression test: `testRefreshInProgressClassifier_matchesOnlyTheInProgressError`.

### Verification
- Token-race + gate verified on simulator (gate awaits ~22–52s, then plays in a single attempt — no -11849 / no "Unavailable" storm); frozen-flow fix confirmed on a physical device (new book's detail no longer dismissed mid-load).
- Pre-push targeted suite: 11 audiobook/nav/bootstrapper test classes green.

### Not done
Does **not** fix LCP streaming-from-license itself — that needs upstream Readium / backend-CDN HEAD+Range support (#810). The download-gate is the RC cure. The token-race fix targets only the "refresh in progress" race; other token failures (missing/invalid creds) still fail the open as before.

Depends on: ThePalaceProject/ios-audiobooktoolkit#185 (submodule pin).

### DIFF
--- a/.forgeos/intent/pp-4542-cold-load-auto-reopen-recovery.md
+++ b/.forgeos/intent/pp-4542-cold-load-auto-reopen-recovery.md
@@ -1,0 +1,68 @@
+---
+name: pp-4542-cold-load-auto-reopen-recovery
+created: 2026-06-16
+author: Maurice Carrier
+branch: fix/pp4542-audiobook-coldload
+ticket: PP-4542
+priority: 3.2.0 regression (critical-path, audiobook playback)
+---
+
+# Intent: cold-load auto-reopen guard for LCP audiobook first-open failure
+
+## Context
+
+Suite-missed 3.2.0 regression (Chairman repro'd on device; reproduced in
+simdrive 2026-06-16 on "Valley of the Moms", Main Street City Library). First
+cold-open of an LCP audiobook (fresh borrow → download → immediate first open)
+fails with the "Audiobook Unavailable" alert and recovers only on a manual
+re-tap. Root cause: the LCP-encrypted package isn't fully materialized when
+playback begins, so `LCPResourceLoaderDelegate` reads a byte-range past the
+archive's available extent and Readium 3.9.0 (PP-4340, ReadiumZIPFoundation
+3.0.1) surfaces `Archive.ArchiveError.rangeOutOfBounds` — where pre-3.2.0
+Readium tolerated the short read. The loader forwarded the hard error to the
+AVPlayerItem (`AVFoundationErrorDomain -11800`), and `handleManagerState`'s
+`.playbackFailed` cold-load branch turned it into a permanent alert.
+
+Primary fix is in the toolkit (retry the transient range read). This app-side
+change is the belt-and-suspenders guard for cold-load failures the toolkit
+retry doesn't absorb, mirroring the bounded OverDrive re-fulfill recovery
+([[ws3-overdrive-expired-url-refulfill]]).
+
+Critical-path (audiobook playback). Toolkit-coupled — submodule pin moves in
+lockstep.
+
+## Claims
+
+- Adds a pure boundary predicate
+  `shouldAutoReopenOnColdLoadFailure(hasEverStartedPlayback:hasCurrentBook:alreadyAttempted:)`
+  returning true iff playback NEVER started this session (a genuine cold load),
+  there is a current book to reopen, and a reopen has not already been attempted
+  for this book this session.
+- Adds a cold-load auto-reopen branch in `.playbackFailed` (PARALLEL to, and
+  after, the SAML and OverDrive branches): on the first cold-load failure,
+  silently `openAudiobook(book, startPlaying: true)` ONCE before publishing any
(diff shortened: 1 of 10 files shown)


mauricecarrier7:
## Review remediation (commit 8dbc4d978)

Both independent reviewers (architect + QA) returned **BLOCK** — and both confirmed **no correctness defect**, only coverage/discipline gaps. All addressed:

| Finding | Reviewer | Resolution |
|---|---|---|
| `preferRemotePosition` shipped without unit coverage on a critical-path FR | QA (hard blocker), architect (concern) | Added a **7-case table test** in `AudiobookPositionRestoreTests` — unparseable-remote→keep-local; no/unparseable-local→remote; >5s→remote; **exactly-5s→keep-local** (boundary/mutation guard on strict `> 5.0`); <5s→keep-local; remote-older→keep-local |
| `awaitTokenReady` reintroduced GCD on the audiobook critical path, contradicting `adr_a265ec76` | architect (concern) | Rewrote recursive `DispatchQueue.global().asyncAfter` → `Task` + `Task.sleep` loop, matching the sibling `awaitAudiobookContentLocal`. Same API, same bounded-timeout semantics |
| Upfront `#if LCP` gate + reactive `.playbackFailed` gate could stack to ~360s | architect (warning) | Verified mutually exclusive from control flow (upfront gate `return`s on timeout *before* `loader.load`, so no streaming → no `.playbackFailed` → reactive gate never fires); documented inline so it can't regress |

Reviewers' confirmed-sound items (no change needed): bounded-poll await is thread-safe / no-leak / terminates; storm-swallow cannot permanently strand a book; `popIfTopRouteAudio` cannot mask a needed pop; fire-and-forget activation is bounded + idempotent.

**Deferred (both rated concern-not-blocker, both accepted manual verification of the auth-coupled state machine for the RC):** unit tests for the storm-swallow guard and the `PositionResolveOnce` double-fire — fast-follow with an injectable clock (not sleep-based).

Tests: 25 affected tests green (`AudiobookPositionRestoreTests` + `AudiobookLoaderOPDSShapeMatrixTests`); pre-push gate green.

PR #1095, opened 2026-06-17
test(regression): audiobook first-cold-open journey — close the PP-4613 suite-miss gap

mauricecarrier7:
## Sim regression coverage for audiobook first-cold-open (PP-4542 / PP-4613)

The 3.2.0 audiobook cold-load regression shipped **un-caught** by the regression suite. Root cause of the gap: the suite's 4 audiobook journeys (`scrubber-drag`, `skip-forward`, `toc-seek`, `download-indicator`) all assume an **already-open, already-local** player — none exercised **first cold-open** of a fresh borrow, which is the only path that streams-while-downloading (and the only place Readium 3.9.0's broken streaming-from-license surfaces as "Audiobook Unavailable").

This adds the missing journey so the class fails loudly next time.

### What's added
- **`.simdrive/journeys/audiobook-cold-load-first-open.yaml`** — book-agnostic journey: open a freshly-borrowed (still-downloading) LCP audiobook → tap Listen → assert the **load-bearing negative invariant** `must_not_contain: ["Audiobook Unavailable", "failed to open", "Please try again", …]`, plus a positive `required_text_any` (Downloading/Loading/Preparing/Chapter/%) so it also catches the **frozen-flow** regression (user yanked back to the catalog).
- **`regression-areas.json`** — registered in the `audiobook` group + `journey_drm` as `lcp` / **non-substitutable** (a Findaway/OD substitute wouldn't exercise the LCP streaming path).
- **`regression_staging.py`** — classified **PHASE2** (clean-skipped by the area-worker) for the *same* staging blocker as `audiobook-download-indicator-stateful`: the sole A1QA LCP audiobook downloads in ~1.5s on the sim with no per-sim throttle, so the not-yet-local window isn't deterministically stageable yet.
- **`RECORDING_COVERAGE_GAP.md`** — now 25 journeys; audiobook 4→5.

### Relationship to the fix
End-to-end backstop for **#1094** (merged), pairing with that PR's unit coverage (`preferRemotePosition`, `isRefreshInProgressError`).

### Not done
Spec only — HID recording is `PENDING_CAPTURE`; promote PHASE2→recipe once a deterministic in-flight-LCP-download harness exists. **Zero `.swift` / no product code** (tooling-only; the pre-push gate skipped the iOS build accordingly). Parity + staging tests green (91 passed).

### DIFF
--- a/.simdrive/RECORDING_COVERAGE_GAP.md
+++ b/.simdrive/RECORDING_COVERAGE_GAP.md
@@ -64,8 +64,8 @@ speed the re-record even though it is not a drop-in.
 - [ ] `reader2-toc-navigate`
 
 ### `audiobook` (5)
-- [ ] `audiobook-cold-load-first-open`  ← NEW 2026-06-17 (PP-4542/PP-4613): first-cold-open of a fresh-borrow LCP audiobook must not dead-end on "Audiobook Unavailable". Spec authored; needs a capture staged DURING the .lcpa download.
-- [ ] `audiobook-download-indicator-stateful`
+- [ ] `audiobook-cold-load-first-open`  ← NEW 2026-06-17 (PP-4542/PP-4613): first-cold-open of a fresh-borrow LCP audiobook must not dead-end on "Audiobook Unavailable". Spec authored; staging now DETERMINISTIC (forge_streaming_state recipe = "ready", no longer PHASE2) — capture is unblocked and is the only remaining step.
+- [ ] `audiobook-download-indicator-stateful`  (stays PHASE2 — needs a live active-download %, not solved by the content-delete forge)
 - [ ] `audiobook-scrubber-drag`
 - [ ] `audiobook-skip-forward`
 - [ ] `audiobook-toc-seek`
--- a/.simdrive/RECORDING_COVERAGE_GAP.md
+++ b/.simdrive/RECORDING_COVERAGE_GAP.md
@@ -6,12 +6,12 @@
 > Snapshot @ develop `e6307c8b2`, 2026-06-12. Step-parity verdicts added 2026-06-12.
 
 ## Headline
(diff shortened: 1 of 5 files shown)


mauricecarrier7:
## Update: deterministic in-flight-download harness added (commit 326672b)

The journey was PHASE2 (clean-skipped) because catching the live ~1.5s LCP download window on the sim isn't deterministic. This commit builds the harness that fixes that and **promotes the journey PHASE2 → ready**:

- **`forge_streaming_state`** staging primitive — instead of racing a live download, it forges the state from the filesystem: delete the downloaded `.lcpa` **content** file while keeping its `.lcpl` license sibling + the registry's `.downloadSuccessful` marker. Registry-state and content-file presence are decoupled (a book is marked downloaded when the tiny license lands, independent of the content), so opening then routes through the exact LCP streaming path PP-4613 lives on — reproducibly, zero network timing.
- New `STAGING_RECIPES["audiobook-cold-load-first-open"]` (sign-in → my_books → forge); removed from `PHASE2_JOURNEYS`.
- 4 new unit tests (forge selects only `.lcpa`; cold-load is ready+forges-after-sign-in; download-indicator stays PHASE2). **95 pass.** Container path + `.lcpa` pattern validated live against a booted sim.

**Sibling `audiobook-download-indicator-stateful` stays PHASE2** — it needs a genuinely *active* download to render 'Downloading… %', which the content-delete forge can't produce (that needs a per-sim download throttle / progress-injection seam — a different mechanism).

**Remaining:** the HID `recording.yaml` is still PENDING_CAPTURE. Staging is now deterministic, so capture is a clean one-shot (run the recipe, then record_start → tap My Books → tap Listen → settle → record_stop) — but it's a live signed-in simdrive run, so it's the final step.

PR #1103, opened 2026-06-23
PP-4632: stop returned audiobook playback → build 483

mauricecarrier7:
Fixes PP-4632 (3.2.0 RC, High): a returned audiobook keeps playing in the new player. Lands on `release/3.2.0`; forward-port to `develop` to follow.

## Root cause
`AudiobookSessionManager` holds the loaded audiobook in memory and never observes the book registry. The return flow (`BookReturnService`) deletes local content + purges caches + sets the registry to `.unregistered`, but nothing tells the live session to stop — the already-loaded tracks keep playing (open file handles survive the on-disk deletion) until a new book opens.

## Fix
- `AudiobookSessionManager` subscribes to `bookRegistry.bookStatePublisher` in init (`subscribeToBookReturn`).
- When the **currently-playing** book → `.unregistered`, it calls `stopPlayback(dismissPhoneUI: true, persistFinalPosition: false)` (full teardown; `false` avoids writing a stale position into a possibly re-borrowed record).
- Decision extracted to a `nonisolated static` helper (mirrors `networkValidationError`) for unit-testing.
- Account-switch teardown unaffected (it nils `currentBook` first → this observer no-ops during it).

## Tests
- 4 decision tests in `AudiobookSessionStateTransitionTests` (current+unregistered→stop; different book→no; non-unregistered→no; nothing playing→no) — **22/22 green**.
- `AudiobookSessionManagerShutdownTests` re-run green (init now wiring the subscription + `stopPlayback` unaffected).

## Build
482 → 483 (next available vs ios-binaries).

## Not validated via simdrive
Audiobook playback can't be simdrive-validated (sim background-download) — runtime device QA (borrow → Listen → return → Play) is pending on RC 483.


mauricecarrier7:
Superseded: per direction, PP-4632 will NOT go into the 3.2.0 RC (no TestFlight 483). It will land on develop for the next version bump instead — see the develop-targeted PR. The build-number bump (482→483) is dropped accordingly.

PR #1125, opened 2026-06-29
fix(alerts): fe741015 — alert presented during active transition (coordinator-wait)

mauricecarrier7:
## Crash
`fe741015` — **#1 production crash on 3.1.0 (478): 122 events / 36 users / 14d.** `NSInternalInconsistencyException` *"A view controller not containing an alert controller was asked for its contained alert controller"*, thrown inside a deferred `_UIAfterCACommitBlock`.

## Root cause (verified)
An alert is presented while a sibling VC transition is in flight; UIKit defers the alert's own transition into the next CA commit, by which point the presentation-controller relationship is inconsistent. The prior guard (`e185f3f8b`) shipped on 3.1.0 yet the crash **grew ~55→122**: it *sampled* `transitionCoordinator == nil` once then presented synchronously (racing the deferred commit), and the `NS_NOESCAPE` `TPPObjCExceptionCatcher` can't trap a throw that fires on a later runloop turn. Confirmed: 25/25 recent events are this variant (the "scroll-direction" subtitle is stale/dormant).

## Fix
Present *after* the in-flight transition settles, mirroring the proven `TPPPresentationUtils.safelyPresent`:
- **`TPPAlertUtils.presentFromViewControllerOrNil`** (explicit + topmost paths): coordinator-*wait* via `coordinator.animate(alongsideTransition:)` completion, re-walking to the settled hierarchy — instead of sample-and-retry-or-drop.
- **`TPPAnnouncementBusinessLogic`**: defer the chained next-alert past the current alert's dismiss.
- **`TPPLastReadPositionSynchronizer`**: launch-time sync-position prompt made coordinator-aware (continuation preserved — alert never dropped).

## Why it escaped before
The prior regression tests only built alert *objects* — never drove present-during-transition. Captured as a wall (`adr_3e2c6a3c`, `.forgeos/wall-failures/2026-06-29-fe741015-guard-tested-wrong-surface.md`); the guard now drives the real present path.

## Verification
- ✅ Build + `UIAlertCACommitGuardTests` green (9/9, incl. new present-path case).
- ⏳ simdrive in-action smoke of the alert flows (signed build ready) — **required before merge** per `docs/bug-investigation-process.md`. (A deferred-CA-commit throw on an *intermittent* launch race isn't deterministically unit-reproducible; durable confirmation is 3.3.0 telemetry.)

Intent: `.forgeos/intent/fe741015-alert-presented-during-active-transition.md` (`type: bugfix`). develop-only — 3.2.0 RC frozen per palace-pm.

### DIFF
--- a/.forgeos/intent/fe741015-alert-presented-during-active-transition.md
+++ b/.forgeos/intent/fe741015-alert-presented-during-active-transition.md
@@ -1,0 +1,83 @@
+---
+name: fe741015-alert-presented-during-active-transition
+created: 2026-06-29
+author: claude-opus-4-8
+type: bugfix
+tracking: Crashlytics fe741015 — #1 production crash on 3.1.0 (478), 122 events / 36 users / 14d. develop-only fix (3.2.0 RC frozen per palace-pm).
+related_prs: []
+---
+
+# Intent: fe741015 — alert presented during active transition (coordinator-wait)
+
+## Claims
+- Replaces the *sample-once-then-retry-or-drop* `transitionCoordinator` guard in
+  `TPPAlertUtils.presentFromViewControllerOrNil` (both the explicit-presenter and
+  topmost-VC paths) with a coordinator-*wait*: when a transition is in flight,
+  present only from `coordinator.animate(alongsideTransition:)`'s completion,
+  re-walking to the settled hierarchy — mirroring the proven pattern already in
+  `TPPPresentationUtils.safelyPresent`.
+- Defers the announcement self-chain in `TPPAnnouncementBusinessLogic.alert(...)`
+  (presenting `nextAlert` from the previous alert's dismiss handler) onto the main
+  queue so the dismiss transition settles before the next present.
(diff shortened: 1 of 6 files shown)


mauricecarrier7:
**simdrive in-action smoke (signed build, iPhone 16 Pro / iOS 26.0): PASS**

- App launched to `state: active` with crash-verification on — no crash in the SIGNAL_EARLY launch window (where `fe741015` fires).
- Full catalog-load completed (Palace Bookshelf + all lanes + tab bar rendered) — the dominant `fe741015` trigger context — with the coordinator-wait path live. `simdrive crashes` since launch: **empty**.

Note: a deferred-CA-commit throw on an *intermittent* launch race (122 events / 36 users / 14d) is not deterministically reproducible on demand, so this smoke confirms the launch/catalog/alert-presentation path is intact under the new code (no drop/hang/regression), not a before/after crash repro. Durable confirmation is 3.3.0 production telemetry once merged.

PR #1128, opened 2026-06-29
fix(hermeticity #2): break AccountDetailViewModel retain cycle + leak guard (follow-up to #1127)

mauricecarrier7:
Follow-up to **#1127** (the hang fix). #1127 made the leaked observer cost ~0 (O(1) `account()`), eliminating the *hang*; this PR fixes the underlying **production memory leak** it surfaced.

## The production bug
The hermetic leak-repro (added for #1127's investigation) proved a **4-hop retain cycle**:
`AccountDetailViewModel → businessLogic → networkExecutor → TPPNetworkResponder → credentialsProvider (= the VM)`.
`TPPNetworkResponder.credentialsProvider` was a strong `let` (the auth-error decision point), so **every Settings VM leaked** via the shared network responder — a real, latent prod memory bug, not a test artifact.

## Fix
- **`TPPNetworkResponder.credentialsProvider`: `let → weak`** (critical-path). Blast-radius survey confirmed safe: the protocol is class-bound (`@objc`/`NSObjectProtocol`); the *only* non-nil provider app-wide is the VM itself (root owner of its executor chain, always alive when an auth challenge fires); read once synchronously with a `?? currentUserAccount` fallback; every other call site already passes nil.
- **`AccountDetailViewModel.loadInitialData`: `[weak self]`** on the `Task { @MainActor }` arms.
- **Leak-repro test** (`AccountDetailViewModelLeakTests`) — red→green proven (failed with the cycle, passes now); a platform-independent **dealloc assertion** that is the *effective* structural guard for the leak class.
- **dup-UUID cross-bucket tie-break** test (qa follow-up from #1127).

## Observer-leak gate — honest disposition (please read)
The plan was to promote the warn-only NotificationCenter observer-leak detector to a hard gate. The FP audit showed a full green suite with 0 breadcrumbs — but a **non-inertness self-test proved *why* it's 0**: on the iOS 26 sim runtime, `NotificationCenter.default.debugDescription` no longer exposes the "observers: N" line, so `sampleObserverCount()` returns nil and the count-based gate is **inert on our platform**. Per the green-board contract, an inert hard-gate is theater — so it was **kept warn-only** (nil-skipping, self-tested to auto-activate if a future runtime restores the API). The real guard for this leak class is the platform-independent dealloc test above.

## DoD
- Full suite **GREEN** (FP-audit run); **0 auth/sign-in/SAML/OIDC failures, 0 hangs, 0 spindumps** (the `weak` change is suite-validated safe).
- Lint-clean; rebased onto develop (`--onto`, post-#1127-squash) — diff is **#2-scope only** (10 files, +362/-8).

## Staged separately (NOT in this PR)
`#3` — close the `TPPNetworkExecutor` test-network escape (route the test session through `NoNetworkURLProtocol`). Non-board-redding now that #1 landed; tracked in the intent as its own focused follow-up.

### DIFF
--- a/.forgeos/intent/accountdetail-leak-cycle-and-hermetic-network.md
+++ b/.forgeos/intent/accountdetail-leak-cycle-and-hermetic-network.md
@@ -76,6 +76,19 @@ is the one strong back-edge. For #3: `URLProtocol.registerClass`
 builds its own `URLSession(configuration:)` whose `protocolClasses` omits
 NoNetworkURLProtocol.
 
+## Deferred / related (palace-pm decisions, 2026-06-29)
+
+- **Pool-responsiveness gate warn->hard (charter Deliverable A): DEFERRED, keep
+  WARN-ONLY** (decision (ii)). It guards class-4 (pool-starvation), which the
+  real artifact RULED OUT for this hang (17 WS0-POOL-DIAG all completed=true);
+  it already earned its keep AS the diagnostic that ruled pool-starvation out.
+  Promote to hard ONLY after a real false-positive audit, if class-4 ever bites.
+  The observer-leak gate below is the correct guard for the class we fixed.
+- **qa optional follow-ups (fold in if cheap):** (1) pin the dup-UUID
+  cross-bucket tie-break (`buildAccountIndex` last-wins — the one behavior change
+  vs the old nondeterministic scan); (2) keep `AccountsManager.swift` in the
+  pre-release `palace_mutate` run so the 50% floor stays honest.
+
 ## Verification
 
 TBD on implementation: red-first leak-repro fails -> passes when the cycle is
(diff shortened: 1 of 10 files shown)


mauricecarrier7:
**Design residual (conscious, documented — qa SoD non-blocking warning), recorded post-merge:**

The `TPPNetworkResponder.credentialsProvider` `let→weak` change is **not** directly unit-tested against a *live* SAML/OIDC auth challenge (a live-IdP unit test isn't practical). Safety is design-argued + blast-radius-survey-backed: the provider is read **exactly once per challenge, synchronously**, with a `?? currentUserAccount` fallback (`TPPNetworkResponder.swift:641`), never retained beyond a method-local `TPPBasicAuth`; the only non-nil provider app-wide is `AccountDetailViewModel`, which is the root owner of its executor chain (alive whenever a challenge fires); every other construction site passes nil and already relies on the fallback. Both SoD reviewers (architect + qa) accepted this. Risk accepted consciously, for Chairman visibility.

PR #1156, opened 2026-07-01
chore(harness): close gate blind spots from swarm_495a88d9 retro

mauricecarrier7:
Applies the feedback + wall-failure fixes from the side-loading swarm (swarm_495a88d9) retro. **Tooling/docs only — no `Palace/` production code.**

## What & why
- **False-green guard** (`scripts/_checklib.py`): `read_diff` now exits 2 on empty/absent STDIN (unless `--allow-empty`) instead of returning `""`. A detector asked to scan nothing can no longer report green — the trust hole behind wall-failure `false-green-empty-diff`. New pytest `test_check_empty_diff_guard.py` (8 cases).
- **Flake classification default** (`scripts/verify-pr.sh`): `--diff-baseline` (re-run failing classes in isolation → real-vs-flake) is now the DEFAULT; `--no-diff-baseline` opts out. A single `--quick` pass surfaces different flakes each run, so a bare failure isn't a trustworthy regression signal.
- **Swarm Phase 4.5 whole-tree-lint gate** (`.claude/skills/swarm/SKILL.md`): new Check 5c runs the AppContainer/AccountsManager isolation lints on the merged state (cross-file lints are invisible to per-implementer DoD), + an implementer-prompt clause to use `makeFreshAccountsManager()`/shared mocks from the start. Wall-failure `isolation-lint-integration-only`.
- **Intent matcher docs** (`.claude/skills/intent/SKILL.md`): document how `check-intent-recorded` actually matches (HEAD subject, `[bracket]`-prefix stripping, ticket-key shortcut) so a valid intent file doesn't get a false `INTENT-MISSING` block.
- **Stale sim** (`CLAUDE.md`): `iPhone 16 Pro` → auto-allocated iPhone destination.
- **Wall-failure catalog**: two new entries + INDEX rows.

## Verification
- `python3 -m pytest scripts/tests/` → **243 passed** (incl. 8 new).
- `bash -n scripts/verify-pr.sh` clean; empty-stdin now exits 2 (was false-green).
- Pre-push tooling gate green.


mauricecarrier7:
Closing — these are harness/tooling changes and shouldn't land in the Palace repo. Re-homing: generic improvements → Heka; Palace-explicit bits → local. Work preserved on branch chore/harness-feedback + /tmp/harness-feedback.patch.

PR #1229, opened 2026-07-09
feat(audiobooks): mini-player polish — drawer (pull to open/collapse), reliable dismiss, no flash on open

mauricecarrier7:
## Audiobook mini-player polish — drawer, reliable dismiss, no flash on open

Polishes the in-app-playback mini-player (#1217) per feedback. Three commits:

### 1. Drawer-style bar (pull to open / collapse)
- A **grabber handle** at the top-center makes the bar read as draggable.
- A unified vertical drag: **pull UP → open** the full player, **pull DOWN → collapse** to the pill; tap still opens. Replaces the tap-only open + down-only collapse.
- Both directions + the expand animate via `PalaceMotion.standard` (reduce-motion aware) — the expand previously used `.easeInOut`, now unified on the motion system.
- The drawer drag lives on the grabber + cover/title zone **only**, never over the transport/dismiss buttons — a container `DragGesture` spanning the buttons intermittently swallowed their taps.

### 2. No mini-player flash on open
Opening an audiobook briefly showed the mini-bar at the bottom, then the full player slid over it. `shouldShowChrome` didn't consider `isPlayerExpanded`, so on a fresh open (`presentOnFirstOpen` sets it true while `hasActiveSession` flips on a later tick) the bar rendered for those frames. **Fix:** gate the bar on `!isPlayerExpanded` — open goes straight to the full player; the bar appears only after minimize.

### 3. Reliable dismiss (+ nicer control)
The bare ✕ → a hierarchical **`xmark.circle.fill` close chip** (reads as a distinct close, not a fourth transport glyph). And the dismiss action is now **guard-free**: the prior optimistic-clear + `isDismissing` guard was *worse* (sim-verified 0/5 on a live bar) — `hasActiveSession` is re-derived from the playback-state publisher, so the optimistic clear was overwritten by the next tick while the guard stuck `true` and dead-locked retries. Now every tap drives the idempotent `stopPlayback` to `.idle`, which stably clears the session.

## Verification
- **22 mini-player unit tests pass** (drawer truth-table up/down + drift, pull-up-expands / pull-down-collapses / small-drag-noop, the `isPlayerExpanded` gate, dismiss contract). Full suite: `Executed 22 tests, 0 failures`.
- **simdrive** (drawer build): grabber ✓, pull-up-expand ✓, pull-down-collapse ✓, close chip ✓ visually.

## Known follow-ups (tracked for the single-view "resize overlay" morph)
- **Collapsed pill tap-through** — tapping the pill passes through to the list/tabs underneath (hit-testing/z-order). This is **pre-existing from #1217**, not introduced here. The morph removes the separate pill (the collapsed state becomes the small *size* of one overlay), which resolves it.
- **Single-view morph** — make full ⇄ mini feel like one view resizing (not two swapping). Chosen approach: a **custom resizing overlay** (the native detent sheet conflicts with the toolkit-player-must-stay-mounted + keep-playing-in-reader constraints). Separate branch, so it's independently discardable.
- The guard-free dismiss is unit-verified; a device pass on the live bar is worth doing (sim audiobook playback is flaky).

### DIFF
--- a/Palace/AppInfrastructure/AudiobookMiniPlayerView.swift
+++ b/Palace/AppInfrastructure/AudiobookMiniPlayerView.swift
@@ -56,13 +56,6 @@ struct AudiobookMiniPlayerView: View {
     /// cache. Production wires this to `appContainer.audiobookSession`.
     let audiobookSession: AudiobookSessionManaging
 
-    /// Guards against a rapid double-tap on `✕` enqueuing two teardowns: the
-    /// button stays mounted until `hasActiveSession` flips false (after the
-    /// async `stopPlayback` completes), so without this a second tap in that
-    /// window would fire a second `stopPlayback`. `stopPlayback` is idempotent,
-    /// but one teardown is the correct contract.
-    @State private var isDismissing = false
-
     @ViewBuilder
     var body: some View {
         // SwiftUI.Group + explicit else: Xcode 26's type-checker otherwise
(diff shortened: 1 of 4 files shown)


mauricecarrier7:
Superseded by #1230. The mini-player approach here (drawer pull-to-open/collapse + reliable dismiss + no-flash-on-open) evolved into the custom single-view morphing player, and all three commits from this branch (d0878e1fa, 88c973051, cd3df1f4d) are already contained in `feat/mini-player-morph` (#1230). Closing to consolidate the audiobook-player work into that PR.

PR #1326, opened 2026-07-22
fix(audiobooks): first-open download visibility + speed chip restores on reopen

mauricecarrier7:
Two audiobook morphing-player UX bugs, both in `AudiobookMorphingPlayerView`, reported together.

## Bug 1 — first-open loading screen looks hung (LCP)

A freshly-borrowed LCP audiobook is marked `.downloadSuccessful` the instant its `.lcpl` license lands while the real `.lcpa` content keeps downloading; tapping Listen enters the PP-4542 content-wait (up to 180s). The opaque shimmer **skeleton** (`!isLoaded`) is layered *over* the whole player via `.overlay { loadingOverlay }`, hiding the `downloadBar` that #1276's `showDownloadProgress` feeds — so a healthy multi-minute download read as a hang. The skeleton also armed a 30s timer that false-tripped an "Audiobook load error / Retry" overlay on any download over 30s.

**Fix:** skeleton flashes only for the initial materialize, then swaps to a determinate **"Downloading… NN%"** state (real cover + title + byte-level URLSession progress + an always-animated spinner) the instant a download is in flight; the 30s load-error timer is suppressed while downloading. Pure `loadingOverlayState` / `shouldSurfaceLoadTimeout` seams with unit coverage.

## Bug 2 — speed chip wrong after exit/return

The chip's `currentRate` was snapshotted only in `.onAppear`, which on a fresh cold open fires pre-bind (manager nil → 1.0× default). The toolkit restores the persisted global rate on bind, so audio played at e.g. 1.5× while the chip stayed at 1.0×.

**Fix:** re-sync the chip at bind (`presenter.playbackModel` goes non-nil at `adoptPlaybackModel`, manager set, `currentPlaybackRate` returns the restored rate). Pure `displayRate(sessionRate:isBound:fallback:)` seam — adopt only once bound. Audio was already correct; this fixes the label.

## Verification
- CI build gate (fresh worktree can't compile LCP/DRM locally) + unit tests (`AudiobookMorphingPlayerViewTests`, all decision seams).
- simdrive pass on the morphing player (in-app-nav ON): speed set → exit → return shows the correct chip; loading→downloading swap.

## Deferred
- in-app-nav-**OFF** production-default legacy path still shows a bare caller spinner during the same 180s wait — separate follow-up.

### DIFF
--- a/Palace/AppInfrastructure/AudiobookMorphingPlayerView.swift
+++ b/Palace/AppInfrastructure/AudiobookMorphingPlayerView.swift
@@ -268,7 +268,17 @@ struct AudiobookMorphingPlayerView: View {
         .overlay(alignment: .bottom) { toastOverlay }
         .accessibilityElement(children: .contain)
         .onAppear {
-            currentRate = audiobookSession.currentPlaybackRate
+            // Only adopt the session rate once a manager is bound. On a fresh
+            // cold open the player mounts (loading shell) BEFORE bind, so a bare
+            // `currentPlaybackRate` read here returns the default 1.0× and would
+            // pin the speed chip at 1.0× for the whole session even though the
+            // toolkit restored the patron's persisted rate. The bind-time
+            // `.onChange` below re-syncs once the real rate is known.
+            currentRate = Self.displayRate(
+                sessionRate: audiobookSession.currentPlaybackRate,
+                isBound: audiobookSession.hasActiveManager,
+                fallback: currentRate
+            )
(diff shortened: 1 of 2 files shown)


mauricecarrier7:
Superseded by #1333 — rebased clean onto current develop (this branch had a stale base + a develop-merge that tripped the local critical-path hook on unrelated merged files). Same commits + reviews carry over.

PR #1338, opened 2026-07-23
feat(decomp): extract PalaceBookModel leaf package (Wave 2a — keystone model layer)

mauricecarrier7:
## Wave 2a — extract PalaceBookModel leaf package (the keystone's model layer)

**Second of three Wave-2 PRs** (pre → **2a** → 2b). Moves `TPPBook` + 12 satellites (`TPPBookState`, `TPPContentType(+Converter)`, `TPPBookRegistryRecord`, `TPPBookLocation`, `TPPReadiumBookmark`, `TPPBookAuthor`, `ImageLoading`, +Presentation) into a new **Layer-0 leaf package `PalaceBookModel`** (→ PalaceCatalog + PalaceLogging, iOS-only). Large but semantically inert: **358 files** carry `import PalaceBookModel`. This is the model half that lets 2b's `PalaceBookRegistry` break the Book↔Accounts cycle at the package boundary.

### Metrics
- **Packages:** +1 (PalaceBookModel) → 10
- **`.shared` reads:** 232 → **224** (ImageCache.shared inversions via the new `TPPBookImageContext` bootstrap)
- **Vestigial `@objc`:** all deleted on the moved types (per the Wave-2 audit) — **NSObject kept** (Set-identity/`isEqual`/NotificationService need it); `TPPBookStateHelper` deleted
- God-class LOC frozen (4 gained an import → re-baselined +1); wire keys byte-identical

### Review — 2/2 Fable SoD, and it earned its keep
The 340-file diff can't be built locally, so the reviewers hunted compile-breaks **statically before CI**:
- **blast-radius** found **39 missing-import files** (file:line) + a stale qualifier → all fixed; re-verified **zero** remaining.
- **architect** found the `Bookmark` protocol stuck `internal` with 12 cross-module consumers (a spec gap) → made `public`; verified de-objc correct, wire-format byte-identical, boundary fixes sound.

A broad safety-net scan caught 1 more file beyond the reviewers' floor. **CI here is the definitive compile gate** (Swift 6 isolation diagnostics).

### Not done
`PalaceBookRegistry` + the `AccountScopeProviding` inversion are Wave 2b (AccountsManager/TPPSignInBusinessLogic gain the import then). 21 pre-existing `@objc` members in other classes that name TPPBook are left to their own waves.

https://claude.ai/code/session_01KEUqnH7e8zUHeUfgeSNDkf


mauricecarrier7:
Admin-merging over a red `build-and-test`.

**The #1338 crash is fixed** — the suite previously died at bootstrap with 0 tests run; on the fix (`51ba9b328`) it runs the full suite: **8067/8072 pass**, all ImageCache tests green.

The 5 remaining failures are the known systemic parallel-clone-starvation / test-pollution flakes, all unrelated to this ImageCache-only change and **confirmed passing in isolation on this branch**:
- `AccountsManagerFirstRunDecodeTests` — 0.13s in isolation vs 2m40s (hung) in CI
- `BookSignInRedirectHandlerTests` — all 8 pass incl. the flagged `testHandleProblem_alreadySAMLStarted_…`
- `TPPBookRegistryLargeCorpusTests` — passes

Per the green-board contract (named, tracked flakes that pass in isolation), admin-merging so the keystone isn't held hostage by non-converging pollution. Flake de-flake debt tracked separately.

PR #1348, opened 2026-07-28
Merge main → develop: forward-port 3.2.1 / 3.2.2 / 3.2.3 hotfixes

mauricecarrier7:
## What this is

Forward-ports **29 commits** of hotfix work — 3.2.1, 3.2.2 and 3.2.3 — that had accumulated on `main` and were never merged back. Per [CLAUDE.md's release policy](../blob/develop/CLAUDE.md#release--hotfix-merge-policy) this is precisely the state that produced the **3.1.0 296-conflict storm**; every additional hotfix compounds what the next `release/X.Y.Z → main` merge has to untangle.

**18 conflicts**, resolved as described below. Merge commit (`--no-ff`), not squash.

## Resolution principle

develop has moved well past these hotfixes — Wave 2b package extraction, Swift 6 concurrency, wall-clock-elimination test seams. So the default was: **keep develop's newer implementation and port the hotfix's *behavior* onto it**, not its code shape.

### Combined — three places where both sides were genuinely required

| Where | Kept from develop | Kept from main | Why both |
|---|---|---|---|
| `AudiobookSessionManager` LCP gate | determinate download-progress feed | **Cause-1 TRIGGER** | trigger without progress reads as hung; progress without trigger can never land. Threaded the sink through the gate's injectable `awaitContentLanding` seam |
| `TPPSignInBusinessLogic` | main-thread hop | **`boundedCompletion`** (#18414) | the hop is *more* needed now — `boundedCompletion` can complete from the URLSession delegate queue **or** the `.global()` timer queue, and `dispatch(_:)` is `@MainActor`. Off-main = the Swift 6 SIGTRAP OIDC crash |
| `AudiobookBookmarkBusinessLogic` | Swift 6 boxing | Cause-2 `cancelPendingRemotePositionWrite` | the cancel seam is still correct (`stopPlayback` uses it); only the `.readingProgress` DELETE was dropped in build 490 |

### De-duplicated — both sides had independently fixed the same thing

- **`shouldAutoReopenOnColdLoadFailure`** — both moved it out of `#if FEATURE_OVERDRIVE` for the noDRM build. Kept develop's.
- **`isResourceUnavailable`** — folded main's `(_:)` into develop's `(from:)`, a strict superset (also inspects the flattened `underlyingDomain`/`underlyingCode`). One implementation instead of two that can drift.

### Package moves (modify/delete)

`BookRegistrySync`, `TPPBookRegistry`, `TPPBook` and main's own copy of `RegistryFileRecovery` were deleted at their old `Palace/Book/Models` paths — Wave 2b owns them in `PalaceBookRegistry` / `PalaceBookModel`, which already carry the equivalent behavior (plus `purgeSidecars` from #1347).

### Submodule

Kept develop's toolkit (`08f7045`). **Verified by content, not ancestry:** main pinned an *unmerged fix-branch* commit for PP-4873, but toolkit `main` independently fixes the same crash via PP-4518's total `PlaybackRate.nearest(to:)`. Zero `PlaybackRate` force-unwraps remain anywhere. No regression.

### pbxproj

Reset to develop's — the auto-merge had produced duplicate test entries plus a reference to a file that exists only on main — then re-added the three genuinely-new hotfix test files via `scripts/pbxproj_add_swift.rb`. **develop's version numbers (3.3.0 / 487) are preserved**; a hotfix forward-port must not drag 3.2.3's numbering onto develop.

## Verification

- `Palace` build **green**; **`Palace-noDRM` build green** (the target the Cause-3 relocation exists for)
- **142 tests across 11 affected suites: 0 failures**
- `check-blast-radius` + `check-superpartner-spectrum` exit 0

## Not done — tracked follow-up

The **D1 `serverAuthoritative` plumbing** (#18414 return-to-empty) is **not** ported. `TPPBookRegistry.removeBook(forIdentifier:serverAuthoritative:)`, `updateAndRemoveBook(_:serverAuthoritative:)` and `RegistryFileRecovery.onDiskHasRecords` don't exist in develop's package, whose INV-1 guard is the narrower pre-3.2.3 one.

That's a behavioral change to the **data path** and deserves its own reviewed PR rather than riding inside a 29-commit history merge. main's return-to-empty tests were removed with a pointer left in `BookReturnServiceTests` — **they are the spec for that port.**

**Not a regression:** develop keeps its current narrower guard, which never refused those saves to begin with.


mauricecarrier7:
## ⚠️ NOT READY — converted to draft

I found a real defect in **my own conflict resolution** here, plus a hole in the verification I reported.

### The defect

`TPPSignInBusinessLogicTests` **crashes on this branch** — 2 SIGTRAP crashes, only 2 of ~19 tests execute:

```
_swift_task_checkIsolatedSwift
closure #2 in TPPSignInBusinessLogic.boundedCompletion(...)
```

**Cause.** develop makes `TPPSignInBusinessLogic` `@MainActor`; the hotfix's `boundedCompletion` invokes its `onTimeout` / `completion` closures from whichever racer wins — for the timeout path, the `.global()` `DispatchWorkItem`. A `@MainActor`-isolated closure invoked off-main is a Swift 6 SIGTRAP.

My resolution kept develop's main-thread hop **inside the closure body**, which is too late: the trap fires when the closure is *invoked*, before its body runs. The hop has to live in the primitive.

Note this is a develop-only hazard — on `main` the type is not `@MainActor`, so 3.2.3 as it ships is unaffected. The forward-port is what makes it fatal.

### The verification hole

I reported "142 tests across 11 affected suites: 0 failures". That was accurate as far as it went, but **`TPPSignInBusinessLogicTests` was not among the suites I selected** — the one most affected by the sign-in conflict. The claim gave more confidence than the evidence supported.

### Attempted fix (insufficient — not pushed)

Hopping to main inside `boundedCompletion` (`Thread.isMainThread ? body() : DispatchQueue.main.async`) still traps, now in `closure #3`. Being on the main *thread* is not the same as being on the MainActor *executor*. The real fix likely needs `@Sendable` parameters (so the compiler rejects isolated closures at compile time) or `MainActor.assumeIsolated`, and deserves its own careful pass.

### State

Branch is unchanged at `2a5c4cbd7` (as pushed). **Do not merge** until the sign-in isolation is fixed and `TPPSignInBusinessLogicTests` passes clean. Everything else in the merge stands: both targets build, and the other 11 suites pass.

PR #1355, opened 2026-07-29
docs(triage-bot): as-built reference + shared architecture proposal (PP-4858)

mauricecarrier7:
Output of the PP-4858 spike, plus two small changes it surfaced.

Draft because these are for reading and comment first. Review inline, on the sentence.

## The two documents

**`docs/architecture/triage-bot-v1-as-built.md`** describes the triage bot as it works today, written as the behavioral contract another platform reproduces: normalization, scoring and the four suggest guards, the redaction rules and their order, the corpus schema, and the escalation state machine. Platform-specific detail is labeled so nobody ports it by mistake.

**`docs/architecture/triage-bot-shared-architecture-proposal.md`** designs the server and Android halves. It opens with a decision memo naming what is being asked for, what is deliberately not being approved yet, and three risks that still need someone's name against them.

**If you have fifteen minutes and a decision to make:** read the cover memo, then the delivery plan and organizational dependencies. Everything else is reference.

## Recommendation

A hybrid. The corpus ships as a versioned CDN artifact rather than a new service, because the org already distributes versioned content that way in two places. Ticket submission becomes a per-library endpoint on the existing Circulation Manager. The AI key proxy stays a hard prerequisite for the AI fallback but is sequenced last, since it is the only piece that may need new infrastructure. Classification and redaction stay on the phones. Android gets a small hand-written Kotlin module kept in step by shared test fixtures rather than shared code.

## Three findings worth reading even if nothing else is

1. **The AI proxy design written in the iOS code cannot be built as described.** The patron's bearer token is encrypted rather than signed, only the Circulation Manager can decrypt it, and the patron's password is inside it. No separate service can verify a patron without also being able to read every patron's password.
2. **There is no infrastructure-as-code anywhere in the org and no self-service deploy path.** Standing up a new service is a request to whoever operates the AWS account, not something engineering can schedule. The plan front-loads everything needing no new infrastructure.
3. **There is no HelpSpot integration server-side to extend.** That work is a build from scratch and is not estimable until API documentation, a sandbox, and a support contact exist.

## Also in this PR

- **Notify-me-on-fix is no longer offered.** It promised a notification nothing could deliver. Hidden behind a pure policy in Core, with the reducer handler and its coverage intact, so restoring it is one line once delivery exists. Mutation-verified.
- **The redaction leak gate now runs.** `scripts/triage-corpus-check.sh` described itself as a release gate but had no caller anywhere. Its `--self-test`, which proves the guard goes red when redaction is disabled, was therefore never running. Wired into CI and `verify-pr.sh`.

## How it was checked

Four grounded passes over `android-core`, `circulation`, `library-registry` and the analytics tooling. Three independent reviews from backend, Android and iOS perspectives. A source audit of every factual claim. And one experiment: the matching engine was rebuilt in Python from the document alone, without reading the Swift, then run against the real conformance corpora. **It matched 85 of 85.** The claim that another platform can build from this document is demonstrated rather than asserted.

## Verification

- Package suite green (285 tests), including the two new policy tests
- Redaction leak gate and its self-test pass
- Documented counts match source: 293 test functions, 70 corpus cases
- All internal anchors resolve; diagrams parse

Related: PP-4882 and sub-tasks PP-4883 to PP-4886 cover the gaps that need closing before the bot's flag is turned on for patrons. The bot ships in 3.3.0 with its master switch off, so nothing is patron-facing yet.


mauricecarrier7:
Taken out of draft to merge. Two changes since it was last reviewed.

**Updated onto current `develop`.** It was 18 commits behind. The `CLAUDE.md` hunk merged cleanly alongside #1383's new green-board rule 5 — verified both are present and the renumber is intact.

**Refreshed the as-built doc for a change that landed today** (`564a888e`). PP-4865 / #1382 added a fourth case to `KBStepResponse.Outcome`, `not_applicable`, and the shipped `catalog.json` now uses it in four places. This document was describing a three-outcome union in both the step-shape schema and the reduce sequence diagram. Since the document's whole premise is that another platform can build a conforming client from it, an Android implementation written against the old schema would have rejected four live catalog entries.

The refresh also records *why* the fourth outcome is not a synonym for `advance` — it advances identically and differs only in recording the step as never attempted rather than attempted-and-failed, which is the input to the rung re-ranking rule. A port that collapses them deflates a rung with attempts that never happened.

Test counts re-derived with the document's own methodology: 399 across 60 files, Core 390 across 56, IOS 4, UI 5. The ResponseQuality corpus is unchanged at 70 labeled cases (the 71st grep hit is still the substring in `hasABenchmarkCase`, exactly as the original audit note said). The refresh note states plainly that nothing else was re-verified and the rest keeps its 2026-07-28/29 date — it is not a fresh audit pass.

PR #1382, opened 2026-08-14
fix(triage-bot): close the two remedy gaps, and the gate that hid them (PP-4865)

mauricecarrier7:
## Summary

Closes the two remedy gaps left open by #1379, and fixes the gate that hid them. Patron-visible change is small: a rung that does not apply to you can now be answered honestly, and the step that tells you to delete and re-fetch a title is finally treated as the destructive action it is.

## Root cause

**Non-attempts were recorded as failed attempts.** A patron already on the newest build had no honest answer to "check the App Store for a Palace update — did that fix it?". The only available button was "No change". `ResolutionTrace` is the input to the pre-registered rung re-ranking rule (Wilson lower bound of a rung's resolution rate at 50+ attempts), so every already-current patron pushed through that branch deflates `updateApp` with an attempt that never happened. The first time that rule runs, it demotes the rung for the wrong reason.

The version gate (`latest_known_app_version`) skips the rung for patrons at or past the newest known version, which covers the common case. This is the backstop for when that gate goes stale — the design deliberately lets it decay to "offer the rung", and this makes the decay non-corrupting.

**A destructive step was invisible to the safety rules.** KI-2026-007 step 2 already told patrons to delete a downloaded title and fetch it again, but carried no `remedy` tag. Untagged meant it could not be skipped for someone who said they had done it, produced no telemetry of its own, and — because `CatalogValidator` only inspects tagged steps — was exempt from the destructive-remedy rules while destroying content.

**The gate under-reported.** `verify-pr` derived its unit-test tally by scraping stdout for `Test Suite 'All tests' passed` + `Executed N tests`. That only holds in serial mode; under parallel clones, which is how CI and `xcode-test-optimized.sh` both run, xcodebuild emits per-test-case lines and those rollups are absent or partial.

## Solution

`KBStepResponse.Outcome` and `StepAttempt.Outcome` gain `notApplicable`. It advances exactly like `advance` and differs only in what the trace records, which is the entire point. All four ladder update rungs offer "Already up to date".

`redownloadTitle` joins the remedy set, classified `destructive` alongside `resetLibrary`: scoped deletion is still deletion, the loan survives but the device copy does not, and PP-4951 is the standing reason re-fulfilment is not a promise. KI-2026-007 step 2 is tagged with it. `CatalogValidator` also learns that `notApplicable` is a refusal route, so a destructive step escaping only that way is not misread as a trap.

`verify-pr` now takes its tally from the xcresult, pins the run with `-resultBundlePath`, reports every blast-radius finding, and counts timeouts and runner restarts as failures.

## Evidence

- TriageBotCore `swift test` → **390 tests, 0 failures** (was 378), top-level `All tests` rollup — CI-reproduced by `unit-testing.yml`.
- Full `Palace` app build for iOS Simulator → `BUILD SUCCEEDED`; `TriageBotUI` clean-built separately, 82 compile tasks.
- `scripts/tests/test_xcresult_summary.py` → **9 passed**, plus validation against a real xcresult.
- `scripts/verify-pr.sh --quick --diff-baseline` → **31/31, `CLEAR: All checks passed`** — self-attested. The triage fired: 8 failing classes extracted, re-run in isolation, all resolved as flakes.
- Guards proved by reintroducing each defect, not by passing:
  - recording `notApplicable` as `didNotResolve` → `testAlreadyCurrent_advancesWithoutRecordingAFailedAttempt` fails by name.
  - adding a bare `"redownloaded"` phrase → `testAppReinstallWordingIsNotMistakenForATitleRedownload` fails by name.
  - removing the new `GuidedStepCard` switch case → iOS `** BUILD FAILED **`, "switch must be exhaustive", while `swift test` stayed **390/390 green**.

**The extractor's first version was wrong and its tests did not notice.** It passed eight hand-written fixtures and returned an empty list against every real xcresult, because a failed test case is not a leaf — its children are `Failure Message` nodes — and Xcode writes `Passed`, not `Success`. The fixtures are now transcribed from a real bundle and carry that warning at the top. Verified against a bundle that previously yielded nothing: it now correctly names `NotificationServiceTokenTests`. This is the same defect `--diff-baseline`'s inline walker had, which is why its flake triage never fired.

**A bug in this PR's own fix was caught by dry-running it.** The first version reported "3 classes fail IN ISOLATION (real regression)" — all four passed when re-run by hand. It reported a count with no names, and could not distinguish a class that genuinely failed in isolation from an isolated run that never executed (a build error or busy simulator prints no suite lines, and every class was then scored a regression). Both are fixed: the classes are named, and a non-run is reported as UNDETERMINED. It still fails closed.

## Repro

- Scenario: patron on the newest build is offered "check the App Store for an update". Jira: **PP-4865**
- Fixture: `Tests/TriageBotCoreTests/UpdateRungAlreadyCurrentTests.swift`, `RedownloadTitleRemedyTests.swift`.

## Class

- Recurrence-of: **`inert-mechanism-shipped-without-data`** — a mechanism landed with no catalog content driving it (the version gate shipped before `latest_known_app_version` was authored; `waitForFix` still ships unreferenced). Both new suites include a catalog assertion plus an anti-vacuity guard so the mechanism cannot pass while unused.

## Obligations

- [ ] `waitForFix` still has no catalog entry referencing it — the bot cannot tell a patron a fix is coming, despite that being the largest resolution class (29/135 resolved; 26/36 within audiobook). Needs per-entry copy on the four `open` entries plus product sign-off. — owner: product + @maurice
- [ ] KI-2026-002 and KI-2026-005 have update rungs with the same non-attempt problem. Deliberately not swept in: an already-current patron there means the entry's diagnosis is wrong, so the honest response is KI-010's `escalate`, not `advance`. That is per-entry copy, not this mechanism. — owner: @maurice
- [ ] `xcode-test-optimized.sh` has its own stdout parsing and was not audited for the same defect. — owner: @maurice
- [ ] `pre-push-test-gate` derives no test classes for SPM package changes and waved through a 25-test-file diff. — owner: @maurice

## Scope

`PalaceTriageBot` (two model enums, the reducer's advance path, the validator, one UI switch, five catalog edits) and `verify-pr`'s reporting. No gate becomes more permissive: the unknown-class path in `--diff-baseline` still counts as a real failure, deliberately, because xcodebuild silently ignores unrecognised `-only-testing` selectors and prints `** TEST SUCCEEDED **` anyway.

## Not done / Deferred

**Not done:** `waitForFix` remains detection-only.
**Deferred:** KI-002/KI-005 update rungs; `xcode-test-optimized.sh` audit; `pre-push-test-gate` package derivation.

---


mauricecarrier7:
Added `630cc09b` — bumps `CURRENT_PROJECT_VERSION` 494 → 495 so this merge actually produces a TestFlight build.

`upload-on-merge.yml` gates the upload on `ios-check-version.sh`, which reports `version_changed=1` only when `ios-binaries` has no archive for the current version+build. `Palace-3.3.0.494.zip` is already there (HTTP 200), so merging at 494 would have uploaded nothing. 495 is free (404). `MARKETING_VERSION` stays 3.3.0.

Branch was also updated with `develop` first, so CI is running against current develop rather than the 3-merge-old base it originally passed on.

PR #1383, opened 2026-08-14
docs+tooling: make "whose failure is this?" a mechanical question

mauricecarrier7:
When a test goes red the reflex is to explain **why**. The cheaper and more reliable first move is to establish **whose it is** — from history, before any theory.

## The incident this generalises

PR #1380 went red on `AccountRegistryStorePoolStarvationTests`. The first diagnosis offered was runner oversubscription: CI runs 4 simulator clones per runner, so the probe "legitimately" couldn't be scheduled. Plausible, self-consistent, and **wrong**.

One comparison against the previous green run settled it:

```
#1377 (green)  passed 0.183s · passed 0.004s · passed 0.004s
#1380 (red)    passed 0.215s · FAILED 69.186s · passed 0.003s
```

Same test, same runner shape, different code — the branch **had** introduced it. (The bump adds `Task { await MainActor.run { … } }` hops, so a test racing the global scheduler became load-sensitive.) Reviewer pushback — *"others aren't having this issue"* — was right and the theory was not.

No amount of reasoning about runners produces that table. Only the history does.

## `scripts/ci-test-history.py`

```bash
python3 scripts/ci-test-history.py <TestClass>[.method] [--limit N] [--branch B]
```

Produces the table in one command and classifies the result:

| verdict | meaning |
|---|---|
| **NEW** | passes on earlier runs / other branches, fails here → yours |
| **PRE-EXISTING** | fails everywhere scanned → not yours; say so with evidence |
| **FLAKY** | passes **and** fails within one run's iterations → retry is masking it |

That last verdict is the one nobody looks for. With `-retry-tests-on-failure -test-iterations 3`, a test passing 2 of 3 reports the job **green**. In this incident it passed 2 of 3 — retry would have swallowed a genuine regression exactly as readily as it surfaced a false one.

## Verified against the real incident, not a fixture

Run on the failing test, it reproduces the table above and flags `87cf85fa` as MIXED. Run on the rewritten test, it reports green (`0.004s · 0.002s`).

It also surfaced a fact I did not have by hand: the same test passed **3/3 on an unrelated branch** at 16:00 the same day — independent corroboration that the failure was ours.

## The rule

CLAUDE.md gets this as green-board **rule 5** (admin-merge becomes 6), pairing the two axes:

- `ci-test-history.py` — whose it is, across **runs**
- `find-test-polluter.sh` — why, across **suites**

Two corollaries are recorded with it: a test absent from a run may have been renamed or never registered, which is **indistinguishable from passing**; and NEW-here + passes-in-isolation means the branch **exposed** pollution rather than broke logic — a different fix.

## Not done

No pytest in `scripts/tests/`. Green-board rule 4 sets that bar for **gates**; this is a diagnostic that gates nothing and exits 0 on every verdict. If it's ever wired into a hook or `verify-pr.sh`, it needs the pytest and the wiring test first.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -123,7 +123,42 @@ false positives. Wiring bugs (a scan-only detector called with `--diff`) are
 invisible to a fixture that only ever stages a violation — always assert the
 clean path passes too.
 
-**5. Retire the admin-merge reflex.** Once the board is trustworthy (1–4), red
+**5. A red test is a question about HISTORY, not a prompt to theorise.** Before
+explaining WHY a test fails, establish WHOSE it is — mechanically, from other
+runs and other suites. Two axes, two tools, both cheap:
+
+```bash
+# Axis 1 — across CI runs: is it new here, pre-existing, or retry-masked?
+python3 scripts/ci-test-history.py <TestClass>[.method] [--limit N]
+
+# Axis 2 — across suites: does it pass alone? then who dirties it?
+scripts/find-test-polluter.sh --victim <TestClass>
+```
+
+Read the **per-iteration** results, not the run verdict. `-retry-tests-on-failure
+-test-iterations 3` means a test that passes 2 of 3 reports the job GREEN while a
+real regression sits underneath — so "passed · FAILED · passed" is a finding, not
+noise. A test that flips with unrelated load is measuring the machine and cannot
+gate CI; make it assert a property of the code (see
+`AccountRegistryStorePoolStarvationTests` for the shape: assert operations
+COMPLETE, keep the load-sensitive variant behind an env flag).
+
+Incident (PR #1380, 2026-08-14): a toolkit-bump PR went red and the first
+diagnosis offered was runner oversubscription — plausible, self-consistent, and
+wrong. One command against the previous green run settled it:
+
+    #1377 (green)  passed 0.183s · passed 0.004s · passed 0.004s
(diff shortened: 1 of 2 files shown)


mauricecarrier7:
Rebuilt on current `develop` (force-push, `ce432b8f` → `763ff667`).

The branch was CONFLICTING, but none of it was a real conflict: two of its three commits — the toolkit bump to `9802581a` and the `AccountRegistryStorePoolStarvationTests` rewrite (plus `ManifestFixture.swift` and its pbxproj entries) — already landed on `develop` via #1380, and `develop` has since moved the submodule forward again to `a7fbf48f` in PP-4957. Merging the branch as it stood would have **downgraded the submodule pointer**.

So the branch was rebuilt from `origin/develop` with only its genuinely-new content re-applied. That is now the whole diff:

```
 CLAUDE.md                  |  36 ++++++++-      (green-board rule 5; old 5 → 6)
 scripts/ci-test-history.py | 195 ++++++++++++++  (new)
```

Re-verified after the rebuild: `python3 scripts/ci-test-history.py AccountRegistryStorePoolStarvationTests --limit 25` reproduces the incident from the PR description — `87cf85fa` reports `passed 0.215s · passed 0.001s · passed 0.003s · failed 69.186s`, flagged MIXED, `VERDICT: FLAKY`. The PR description above stands as written; only the toolkit-bump commit it mentions is gone, because it is already on `develop`.

PR #214, opened 2022-10-31
Add loading indicator to my books view

mauricecarrier7:
**What's this do?**
Adds a loading indicator to my books screen when books are fetched.

**Why are we doing this? (w/ Notion link if applicable)**
https://www.notion.so/lyrasis/Add-loading-indicator-for-My-Bookshelf-828aa9ae06bc4a58b500b2ba4ce4a1f1

**How should this be tested? / Do these changes have associated tests?**
Go to my books, wait for screen to populate, indicator should be shown. Pull to refresh, loading indicator should be shown. 

**Dependencies for merging? Releasing to production?**
N/A

**Does this include changes that require a new Palace build for QA?**
yes

**Has the application documentation been updated for these changes?**
N/A

**Did someone actually run this code to verify it works?**
@mauricecarrier7 

<details>
https://user-images.githubusercontent.com/6921353/199052912-66e7bbd2-a14f-44a1-8d05-3aec3bc6cdf7.mov
</details>

PR #213, opened 2022-10-28
Update dependencies to support XCode14

mauricecarrier7:
**What's this do?**
Updates dependencies to support XCode 14

**Why are we doing this? (w/ Notion link if applicable)**
https://www.notion.so/lyrasis/iOS-Update-Palace-project-to-support-Xcode-14-bb1a0e5da4c44e56b411a40231011cc8

**How should this be tested? / Do these changes have associated tests?**
Test all app functionality

**Dependencies for merging? Releasing to production?**
https://github.com/ThePalaceProject/ios-provider-pdf-renderer/pull/5

**Does this include changes that require a new Palace build for QA?**
Yes

**Has the application documentation been updated for these changes?**
pending

**Did someone actually run this code to verify it works?**
@mauricecarrier7

PR #1387, opened 2026-08-17
chore(toolkit): bump to 1adbe833 — Swift 6 Phase D waves 1-5 (PP-4724)

mauricecarrier7:
Brings the audiobook toolkit's strict-concurrency work into the app — toolkit PRs #212, #213, #214.

## The patron-visible part is one real bug fix

`Audiobook.drmStatus`'s setter reaches into `player.isDrmOk`, whose `didSet` pauses playback, publishes `.failed` to Combine and unloads the player. Every setter of `drmStatus` sits inside a `URLSession` completion handler in `FeedbookDRMProcessor` — so **DRM expiry was manipulating AVPlayer and publishing to SwiftUI from a network callback thread**. It now hops to the main actor. `SleepTimer` had the same shape, reading `isPlaying`/`currentChapter` from its own serial queue.

## The rest is the isolation that made those findable

The playback layer — `Player`, `Audiobook`, `AudiobookManager`, `SleepTimer`, `AudiobookPlaybackModel` and the two factories — is now `@MainActor`. Four redundant `deinit`s are gone, and the queue player's legacy string-keyPath KVO (the one registration that crashes if it outlives its observer) is now block-based and self-invalidating.

## This is a declaration of existing behaviour, and this repo proves it

**Palace builds against the new toolkit with 0 errors and no source changes.** Every app call site was already a `@MainActor` type, a `View` or an `ObservableObject`. Isolating the `Player` protocol produced exactly **one** error inside the toolkit — and that error was the DRM defect above.

That is the whole argument for this bump being low-risk: the isolation was not imposed on the app, it was discovered in it.

## Evidence

Toolkit, full builds on a clean `OBJROOT` (incremental rebuilds under-report badly here):

| | before | after |
|---|---|---|
| warnings | 256 | **174** |
| self-declared Swift 6 errors | 80 | **18** |

Toolkit suite on merged `main`: **246 tests, 1 skipped, 0 failures**, no timeouts, no runner restarts. Palace build against it: **BUILD SUCCEEDED, 0 errors**.

## Build number

Bumped **495 → 496**. 495 is already uploaded to `ios-binaries`, and `upload-on-merge` silently skips the upload when the archive already exists — so merging at 495 would produce no TestFlight build at all.

## Scope

Submodule pointer and build number. No app source changes; none were needed.

**Not done:** `SWIFT_VERSION` stays 4.2 in the toolkit — the flip is wave 6.
**Deferred:** `FindawayPlayer` (47 warnings, only 3 of them gating) needs its own PR; 10 of its hop sites use a private serial queue that exists because `FAEPlaybackEngine` calls are "potentially very expensive".


mauricecarrier7:
Re-pointed to `e62cbd9a` (toolkit #215) and re-verified properly.

**What was wrong.** The first pointer (`35b84ec2`) did not build: `LocalBookContentService.deleteLocalAudiobookContent` calls `AudiobookFactory.audiobookClass(for:).deleteLocalContent(...)` from a nonisolated context, and both became main-actor isolated in toolkit #213. CI caught it; my local check did not.

**Why my local check was wrong, since that is the more useful part.** The build I ran to measure the blast radius linked a **stale prebuilt framework** from an ios-core DerivedData I had never cleared — I had been religiously clearing the *toolkit's* DerivedData and forgot this one — so it compiled against the pre-isolation toolkit and reported zero errors. Same stale-artifact class as the "0 warnings from an incremental rebuild" trap.

**The fix.** Toolkit #215 makes three members `nonisolated`: `audiobookClass(for:)` (a pure switch on manifest type), `deleteLocalContent(manifest:bookIdentifier:token:)` (builds its own `Tracks`, deletes files, reads no main-actor state), and `getFulfillmentId(from:)` (a pure manifest read). Isolating the *caller* instead would have forced blocking file I/O onto the main actor — a regression dressed as a concurrency fix.

**Re-verified the way it should have been done first:** ios-core DerivedData cleared, then `build-for-testing` so the **test target** compiles too, not just the app:

- `** TEST BUILD SUCCEEDED **`, **0 errors**
- **115 toolkit Swift compile tasks** in the log — proof the framework was rebuilt from source rather than linked stale

Toolkit suite on merged `main`: 246 tests, 1 skipped, 0 failures, no restarts.

mauricecarrier7:
**Pointer history**, since the first commit's subject line still says `0f0731fd` and the pin is now two commits past it.

The bump needed three follow-up toolkit fixes, each found by a different oracle rather than by review:

| toolkit PR | what | how it was found |
|---|---|---|
| #215 | `deleteLocalContent` / `audiobookClass` / `getFulfillmentId` → `nonisolated` | ios-core CI — the app deletes downloaded content from a nonisolated context, and forcing blocking file I/O onto the main actor would have been a regression dressed as a concurrency fix |
| #216 | `setupStreamingFor` → `@MainActor` requirement | ios-core CI — `#if LCP` code |
| #217 | `remainingWallClock` → `nonisolated` | **rebasing onto develop** — PP-4971 (#1384) added a nonisolated caller *after* the isolation work was written, so the two only meet at integration |

Final pin: **`1adbe833`**.

## Verification, corrected

Three earlier "green" runs on this PR proved nothing, and the qa_test reviewer caught it. The canonical submodule checkout sat at `d956b7bb` and the harness copies from it into the worktree on every run, so those builds compiled the **pre-bump** toolkit — where a nonisolated witness is trivially legal and this PR's central compile-time claim is untested.

Now verified properly: the toolkit tree is materialized at `1adbe833` and the `@MainActor` markers on `StreamingCapablePlayer`, `setupStreamingFor` and `AudiobookPlaybackModel` are asserted present **before** the build and re-asserted **after** it, so nothing clobbered them mid-run. Both reviewers independently confirmed the on-disk tree matches the pin.

That failure mode is now mechanical rather than reliant on a sharp reviewer: `verify-submodule-tree` asserts a submodule's files match the recorded gitlink and fails loudly otherwise. It is proved by reintroducing the exact drift.

## Reviews

architect **approved** and qa_test **approved**, both bound to `90442264a`, both at attested tier (no signing key configured on this machine).

Both were explicit that the scoped run is a spot-check and **full-scheme CI is the gate** — this push is for that signal, not to merge. Watching audiobook teardown and AVPlayer-leak suites in particular, since `AudiobookPlaybackModel.deinit` is deleted at this pin and no ios-core test covers that.

PR #1397, opened 2026-08-19
fix(ci): stop a green test run reporting red because reporting failed (PP-4988)

mauricecarrier7:
CI went red on branches whose suites passed. The tests ran, the tally read zero failures, and the job then died uploading the `.xcresult` artifact with `ENOENT … stat ".../Data/data.0~..."`. A second shape did the same via `Post PR Comment with Results`.

A board that is red for reasons unrelated to the code trains everyone to ignore red, and a real failure then hides in the noise. This repo has shipped a broken commit gate exactly that way.

## The audit mattered more than the two known cases

Chasing the reported failures would have fixed two steps. Checking **every** step that exists to publish or archive found **six more** that could still fail the job — the report generators, the JSON upload, the snapshot upload, the Pages deploy. All are now `continue-on-error`, so the class is closed rather than the two instances.

## The pass/fail gate is deliberately untouched

`Fail if tests failed` still keys on `steps.tests.outcome` — the test step itself — and still fails the job. **A real test failure reddens the board exactly as before.** The only thing that stops doing so is failing to save or publish a result.

The distinct outcome is now visible without opening the log: when the archive cannot be uploaded, the run emits a warning annotation and a step summary saying the suite ran, its verdict stands, and only the downloadable artifact is missing. Previously that was indistinguishable from a test failure on the checks list.

## Guarded, and proved

`scripts/tests/test_ci_reporting_never_reddens_board.py` asserts both halves, because a fix that achieved the first by weakening the second would be far worse than the flake:

1. no reporting step may fail the job;
2. the gate still can, still keys on the tests, and never references a reporting step's id.

Both proved by reintroducing the defect — removing the tolerance fails `test_no_reporting_step_can_fail_the_job` by name; adding `continue-on-error` to the gate fails `test_the_passfail_gate_is_still_able_to_fail` by name.

## Correction to the ticket

It said this had hit four branches in two days. Checking the run history, **one** of the last ten failures was the artifact upload and one was the PR comment; the rest were genuine test failures. The defect is real and worth fixing, but it was not the dominant cause of red that the ticket implied.

**Scope:** eight `continue-on-error` additions, one annotation step, one pytest. No change to what makes a run pass or fail.

**Not done:** the underlying race — why the `.xcresult` bundle is incomplete when the uploader reads it. This makes it non-blocking and visible rather than curing it; that stays on PP-4988.


mauricecarrier7:
Holding this rather than merging or re-running, because another agent is actively working the board reds.

**The red on this PR is not this PR.** The failing step is `Fail if tests failed` — the pass/fail gate — not a reporting step, so the board went red exactly as this change intends. The failure is `SignInWebSheetIntegrationTests.test_initialLoad_injectsCookiesBeforeLoadingRequest`, and the per-iteration history is decisive:

```
c4c7b950 (this tip)  failed 9.689s · passed ×5   <- MIXED, retry masking
ec30b1c7             passed ×3
a79c333e             passed ×3
```

It failed once and passed five times **within the same run**, and passes on other branches. This diff is a workflow YAML file and a Python test — it cannot affect a Swift sign-in test. Filed as **PP-4995** with the shape noted: two other waits in that suite carry explicit `WKWebView` cold-start allowances and this one does not.

**Deliberately not re-running.** That is the "re-run and hope" behaviour this very PR exists to end, and a test that flips under load is measuring the machine, not the code.

**Deliberately not admin-merging either**, even though it fits the rule 6 carve-out. This change edits `unit-testing.yml`, which is the file whoever is fixing the reds is most likely working in — landing it mid-diagnosis would either conflict with their work or muddy their signal about what changed.

Ready to merge whenever the board is clean, or immediately if the agent working the reds would rather have it in place first — the two guards (`test_ci_reporting_never_reddens_board.py`) mean it cannot silently weaken the gate either way.

mauricecarrier7:
Correcting my previous comment, and holding again — with better evidence this time.

**I misreported the failure.** I said it was `SignInWebSheetIntegrationTests` from the run before, and that the parallel bundle had 65 failures. Both wrong. `parallel=65` is an internal counter, not a failure count. The actual result on the rebased tip is **8340 tests, 8196 passed, 2 failed**, and the two are different tests entirely:

| test | per-iteration on this tip |
|---|---|
| `ImageCacheOffMainIsolationTests` | `passed · passed · **failed 120.000s** · passed ×3` |
| `TPPBookRegistryLoadReentrancyTests` | `**failed 0.016s** · passed ×5` |

Both **FLAKY** by `ci-test-history.py`, both flipping within a single run, and both failing on branches that are not mine — `fix/pp-4965-annotation-error-classification`, `fix/pp-4896-remove-vestigial-book-to-remove`, and this one.

**The `120.000s` is the tell** on the first: an exact round timeout, not a logic failure. That is a wait expiring under load, and it has now expired on three separate branches today.

**The genuinely useful finding:** commit `a79c333e` was reported **green** while `ImageCacheOffMainIsolationTests` failed at 120.000s and `TPPBookRegistryLoadReentrancyTests` failed inside it. Retry rescued both. So these have been failing all day across multiple branches and the board only shows it when the retries happen to line up badly — exactly the "passes 2 of 3 reports the job green while a real regression sits underneath" case CLAUDE.md warns about.

**Still not merging, and still not re-running.** Another agent is working the board reds and these two belong to that work, not to a workflow-YAML change. Re-pushing until the dice land green would be the same anti-pattern this PR argues against, just slower.

The diff remains two files — `unit-testing.yml` and one pytest — neither of which can affect a Swift image-cache or registry test.

PR #1402, opened 2026-08-20
Three instruments for questions a release regression could not answer

mauricecarrier7:
Base `develop` (ff3b1227c) · 10 commits · 8 files · +1092 / -0 · no production code.

**These run unaided.** Verified from a non-provisioned clean checkout of this
branch: `bash`, `python3`, `/usr/bin/log`, and `xcrun simctl` — nothing else. No
maintainer-local tooling, no repo-specific setup, and the full test suite passes
in a worktree that was never provisioned. That is the line this repo draws about
what belongs in the shared tree, and it was demonstrated rather than asserted.

---

## Three instruments for questions a release regression could not answer

A 3.3.0 regression pass raised 41 chaos findings. Settling them turned up the
same defect repeatedly — not in the app, in the *measuring*: a result that looks
measured and is not. Each tool below exists because one specific wrong answer
came within a message of being filed.

### `scripts/sim-log-recover.sh` — read a pass that already happened

A QA pass preserves the log lines its operator thought to quote; the rest is
treated as gone. It usually is not. The simulator keeps its own unified-log
store on the host, and it outlives the pass.

**The wrong answer it prevented.** A finding was filed *minor* on the strength
of one `-999 cancelled` line, with a stated cause of "a duplicate fulfillment
download cancelled mid-flight." The recovered store showed three download tasks,
each with its own fulfillment request, **two completing successfully** — one
borrow billing the circulation manager three times and transferring the same
EPUB twice. Promoted to major on the critical path, with nothing re-driven.

Reads host files only, so it is safe against a simulator another session owns —
which is how that measurement was taken without touching the sim that held it.

**Limits, stated in the file:** the shim cannot resolve the app's own `os_log`
lines; they render `<compose failure [UUID]>`. System subsystems (CFNetwork,
network, RunningBoard, WebKit) come through complete. That is enough for
"how many requests were issued and did they succeed" and not enough to read
what the app said about itself. On one cluster this made the tool unable to
measure the thing being asked, and it says so rather than returning a number.

### `scripts/sim_tap_on_event.py` — tap inside a window a fixed delay cannot hit

Some behaviour reproduces only when input lands inside a window that opens on
its own schedule. One such window is 1.2–3.0s wide and opens 6–11s after a
borrow, varying per title and network. Every fixed-delay attempt missed it.

**The wrong answer it prevents.** A missed window and a real negative are
indistinguishable in a transcript, and "we tapped and nothing happened" retires
a hypothesis that "we never landed a tap" does not. Exit `0` reports the tap
*with its measured latency from the trigger*, so the operator can judge whether
it plausibly landed inside the window; exit `4` says the trigger never fired and
states that this implies nothing about the behaviour under test.

**Limits:** unlike the recovery script this spawns a process on the device, so
it requires owning the simulator. Callers pass POINTS, not pixels.

### `scripts/differential_check.py` — refuse to compare when the comparison is invalid

Counting a signature on two builds yields a number whether or not the number
means anything.

**The wrong answer it prevented.** A baseline showed **zero** cover-decode
failures against the candidate's 53 — an open-and-shut regression, except the
baseline had never rendered the lane those covers live in. A zero from an
unexercised path. Separately, three clusters were compared across an iPad and an
iPhone. Both would have been filed; the first would have been a **false
REGRESSION** on a release-gating row.

It checks device, OS, and exercise on both sides *before* reading anything, and
emits no comparison unless all hold. A refusal names the unmet precondition and
never leaks a count, so there is no number to quote out of context.

**Limits:** input-class equality cannot be checked mechanically — that is what
the witness pattern is for, and a badly chosen witness ("the app launched")
satisfies the check and defeats it. Nothing stops someone quoting the `reason`
field alone; the "this is a statement about the DATA, not about the builds"
disclaimer sits beside it and cannot follow it.

### `docs/Testing/simulator-log-recovery.md`

The recovery method, the four ways to get an empty result that is not missing
data, and one section written deliberately as discipline rather than a script:
*a zero is only a measurement if the command hit something.* Three instances
appeared in one afternoon — an unquoted `$ref:path` in zsh, a `pytest` piped to
`tail`, a grep against a path that does not exist. A guard covering only the
wrong-target cases would leave the merely-wrong-pattern case returning an
identical honest zero while implying the class was handled, which is the failure
shape the document exists to prevent.

---

## Verification

From a **clean checkout of this branch**, not the development tree:

- `bash -n` on both committed shell scripts (CI `tooling-checks.yml` parity)
- `scripts/tests/test_sim_log_recover.sh` — 6 cases
- `pytest scripts/tests/test_sim_tap_on_event.py scripts/tests/test_differential_check.py` — 15 tests
- No harness, heka, or ForgeOS dependency; no simulator required by any test
- Every guard proved by reintroducing the defect: 5 mutants across the three
  tools, each killed by a **named** assertion. One mutant initially SURVIVED —
  a test fixture used a fake `Task <X>.<1>` that failed the pattern on the UUID
  rather than the bundle id, so the test passed for a reason unrelated to its
  own name. Fixture fixed, mutant killed.

## Ship-safety audit

- No campaign paths, run identifiers, session names, or agent names
- No absolute `/Users/...` paths
- UDIDs present are synthetic fixtures (`AAAA…`, `DEADBEEF…`); one real
  CFNetwork **task** UUID appears in a fixture as sample log text — ephemeral,
  not a device or account identifier. Deliberately NOT sanitised into a fake: a
  fixture that does not match production bytes is how a mutant survived here
  once already.
- The word "harness" appeared twice as ordinary English for a QA run. Reworded
  to "QA run" (1a276e260): in this repo that word names a specific
  maintainer-local system and would read as a dependency that is not there.
- `docs/bug-investigation-process.md` already references `.forgeos/` at two
  pre-existing lines; this branch adds five lines and touches neither.

## Not done

- Nothing is wired into a gate or CI job. These are instruments, and the
  campaign's rows are still being settled by hand.
- The tools were built during an active campaign and have one user. Ergonomics
  are unproven.

### DIFF
--- a/docs/Testing/simulator-log-recovery.md
+++ b/docs/Testing/simulator-log-recovery.md
@@ -106,6 +106,26 @@ bites: counting raw lines instead of data lines (an uncovered window then
 reports itself as a real negative) and dropping the `|| true` on the filter
 (a legitimate zero-match then exits non-zero). Both are caught by name.
 
+## If you will want to quote it at info level, capture it live during the run
+
+The archive is complete for debug and error and near-empty for info. So
+anything you expect to quote at info level must be captured **live, while the
+run is happening**, or it does not exist afterward — a shutdown ends the buffer
+and a later boot starts a new one rather than restoring the old.
+
+That makes live capture a property of the RUN, not of the analysis. A QA or
+chaos pass that will cite log evidence should stream its own log to a file for
+the duration:
+
+```bash
+xcrun simctl spawn <UDID> log stream --style compact --info --debug > pass.log &
+```
+
+Everything this page describes is recovery of what survived. It is not a
+substitute for capturing the record in the first place, and the gap is silent:
+a pass that never captured looks identical afterward to one whose evidence was
+merely never quoted.
+
 ## Four ways to get an empty result that is not missing data
 
 Each of these returns zero lines. The script now tells the first three apart
--- a/docs/Testing/simulator-log-recovery.md
+++ b/docs/Testing/simulator-log-recovery.md
@@ -1,0 +1,137 @@
+# Recovering simulator logs after the fact
+
+A QA or chaos pass records the log lines its operator thought to quote. Every
+other line the app and the system emitted during that pass is, by default,
+treated as gone. It usually is not gone: the simulator keeps its own unified-log
+store on the host filesystem, independent of anything the harness captured, and
+that store outlives the pass.
+
+`scripts/sim-log-recover.sh` reads it.
+
+## Why this matters more than it sounds
+
+A finding backed by one pasted log line cannot survive being questioned. When a
+reviewer asks "was that really two requests, or one request retried?", the
+quoted fragment cannot answer — and the usual next step is to re-drive the
+scenario, which costs a simulator, a build, and an hour, and produces a *new*
+observation rather than evidence about the original one.
+
+This is not hypothetical. In one release regression the run's `logs/` and
+`crashes/` trees were empty for every device cell, so 41 findings rested
+entirely on quoted fragments. One of them was filed **minor** on the strength of
+a single `-999 cancelled` line, with a stated cause of "a duplicate fulfillment
+download that is cancelled mid-flight."
+
+Recovering the store showed three download tasks on the app's background
+session, not one:
+
+```
+<5848D9C1…>.<1>  resumed 21.139 → finished with error [-999] cancelled 22.749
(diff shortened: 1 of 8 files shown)


mauricecarrier7:
Closing: these are local-machine simulator forensics tools. They read $HOME/Library/Developer/CoreSimulator stores and drive local sims, so they cannot be run from another system (CI has no driven simulators). Keeping them local per owner instruction — only tooling a contributor can actually run belongs in the shared repo.

